PR #6, opened 2017-08-01
upgrade to Rails 5

xithan:
breaking changes from from 4.2 to 5.1:
- dirty attributes behave differently in after save callbacks
- parameters don't inherit from hash
- controller test api


card/lib/card/migration.rb
```diff
@@ -1,6 +1,6 @@
 # -*- encoding : utf-8 -*-
 
-class Card::Migration < ActiveRecord::Migration
+class Card::Migration < ActiveRecord::Migration[4.2]
```
ethn: Is this a typo, or does that actually mean something?

card/lib/application_job.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
+end
```
ethn: I would have guessed that either (a) this is defined elsewhere and thus not needed or (b) it isn't and thus we can continue inheriting directly from ActiveJob::Base, but I'm probably missing something?

card/mod/core/set/all/actify.rb
```diff
@@ -128,3 +128,11 @@ def with_transaction_returning_status
 def success
   Env.success(cardname)
 end
+
+def name_before_act
+  name_before_last_save || name_was
+end
```
ethn: perhaps a comment here?  guessing behavior changed to something a bit more surprising?

card/spec/lib/card/i18n_spec.rb
```diff
@@ -29,6 +29,7 @@ def card_gem_root
   end
 
   it "does not have unused keys" do
+    pending
     unless use_i18_tasks
       skip "upgrade to Ruby 2.1+ and install i18n-tasks gem"
```
ethn: should always be 2.2.2+ now, right?

card/lib/card/migration.rb
xithan: no, that means something and it doesn't work without that

xithan: it means we are using the rails 4.2 version of AR::Migration

card/lib/application_job.rb
xithan: I'm following the upgrade guide there with introducing that class and since CoC is one of rails' core principles and a new convention for 5.2 says always inherit from Application... and never from ...::Base I thought it's better to follow that pattern.

card/mod/core/set/all/actify.rb
xithan: I was thinking if the names are not obvious enough you can look it up in the rails documentation, but I'll write a comment for you.

PR #9, opened 2017-08-04
bootstrap carousel helper

xithan:


card/mod/bootstrap/lib/bootstrap/component.rb
```diff
@@ -91,11 +101,27 @@ def add_content content
...
+          wrapper = wrappers.shift
+          if wrapper.is_a? Symbol
+            send wrapper, &content_block
+          else
+            instance_exec(content, &wrappers.shift)
+          end
+        end
+      else
+        add_content content
+      end
```
ethn: getting to be a pretty beefy method.  Can we break it into smaller parts?

xithan: The whole bootstrap builder API needs refactoring. The wrap stuff doesn't work and it's way too messy and complicated to be maintable.

PR #12, opened 2017-08-09
Pov

ethn:
pov names (in progress)


xithan:
What does pov mean?

ethn:
"point of view".  I'm not in love with it.

I'm working on the problem that `relative` is overloaded, sometimes it means a name with `_left`, `_right`, etc.  Other times it means a name as seen from a certain perspective.  So `name.relative_name.relative?` is often false, which is weird, no?

I think maybe `pov_name` should be `name_from`.  (Though that doesn't really work as the replacement for card.relative_name. Sigh)

card/lib/card/format/nest/subformat.rb
```diff
@@ -3,11 +3,24 @@ class Format
...
         def subformat subcard
-          subcard = Card.fetch(subcard, new: {}) if subcard.is_a?(String)
-          self.class.new subcard, root: @root, parent: self, depth: @depth + 1,
-                                  form: @form,
-                                  mode: @mode,
-                                  context_names: @context_names
+          subcard = Card.fetch(subcard, new: {}) unless subcard.is_a?(Card)
+          self.class.new subcard, parent: self, format_class: self.class, form: @form
+        end
+
```
ethn: This is all the new lazy inheritance handling

card/mod/basic_formats/set/all/base.rb
```diff
@@ -6,16 +6,19 @@ def show view, args
...
   end
 
   view(:key,      closed: true, perms: :none) { card.key }
   view(:linkname, closed: true, perms: :none) { card.cardname.url_key }
   view(:url,      closed: true, perms: :none) { card_url _render_linkname }
 
   view :title, closed: true, perms: :none do
-    raw_title = voo.title || card.name
-    voo.variant ? raw_title.to_name.vary(voo.variant) : raw_title
+    name_variant title_in_context(voo.title)
```
ethn: This is a bit of a shift, in which we assume that the `title` is rendered in context, and the `name` is the full name.  I'm inclined to say it makes sense to use that as a new default (but to make contextualization variants possible, perhaps via `variants`.

ethn:
I think this is ready to merge, though I'm sure I'll need to make some updates in WikiRate.  I settled on `name_from` and `from` as the main API to what used to be `relative_name` or `to_show`.

PR #14, opened 2017-08-24
Formnames

ethn:
main changes:

1. better names for basic form methods and variables 
2. field_prefices are always relative.
3. some removed API (because mod developers don't have to think about relative form names any more)

(more inline)

### DIFF
--- a/card/mod/standard/set/all/rich_html/form.rb
+++ b/card/mod/standard/set/all/rich_html/form.rb
@@ -20,7 +20,7 @@ format :html do
 
   view :edit_in_form, cache: :never, perms: :update, tags: :unknown_ok do
     reset_form
-    @edit_in_form = true
+    @in_multi_card_editor = true
     edit_slot
   end
 
@@ -101,7 +101,7 @@ format :html do
 
   # test: are we already within a multi-card form?
   def in_multi_card_editor?
-    @edit_in_form.present?
+    @in_multi_card_editor.present?
   end
 
   def nests_editor?


card/mod/core/set/all/name.rb
```diff
@@ -97,15 +97,6 @@ def raw_name
...
 
-def relative_name context_name=nil
-  context_name ||= @supercard.name if @supercard
-  name.name_from context_name
-end
-
-def absolute_name context_name=nil
-  context_name ||= @supercard.name if @supercard
-  name.absolute_name context_name
-end
```
ethn: REMOVING API (from Card).  I checked WikiRate to try to find any instances where it is used but could not find it.

I got into this area a while back because I was trying to clean up the contextualization stuff for view caching. Most of the name context stuff was discussed and merged in long ago, but I couldn't get rid of these calls until I upgraded the formname handling. 

(by the way, I expect there will be some name context stuff that will be found and fixed this next week, but the work table is much cleaner now)

card/mod/core/set/all/subcards.rb
```diff
@@ -67,7 +67,7 @@ def unfilled?
   subcards.each do |subcard|
     subcard.errors.each do |field, err|
       err = "#{field} #{err}" unless %i[content abort].member? field
-      errors.add subcard.relative_name.s, err
+      errors.add subcard.name.from(name), err
```
ethn: moves from Card to Card::Name

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -19,8 +19,8 @@
   end
 
   view :edit_in_form, cache: :never, perms: :update, tags: :unknown_ok do
-    @form = form_for_multi
-    @within_multi_edit = true
+    reset_form
+    @edit_in_form = true
```
ethn: I find this nomenclature much clearer.  line 22 was actually just resetting the form, even before.  It _looked_ like it was creating a new kind of form (a "form_for_multi"), but all forms were generated with that method.  That confused me.

Similarly, I think just changing this variable to @edit_in_form makes this easier to hold in your head.

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -140,30 +140,50 @@ def process_edit_fields fields
...
-    instantiate_builder(form_for_multi_input_name, card, {})
+  # If you use subfield cards to render a form for a new card
+  # then the subfield cards should be created on the new card not the existing
+  # card that build the form
+
+
+  def form
+    @form ||= begin
+      @form_root = true unless parent&.form_root
+      instantiate_builder(form_prefix, card, {})
```
ethn: I prefer `form_prefix` to `input name`, because the complete index name has the field name in it, and I think it might have a slightly broader use than just inputs (?).  But we could change this, eg, to `input_prefix` or `input_name_prefix`

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -227,16 +236,6 @@ def editor_wrap type=nil
...
-  def with_relative_names_in_form
-    @relative_names_in_form = true
-    result = yield
-    @relative_names_in_form = nil
-    result
-  end
-
-  def relative_names_in_form?
-    @relative_names_in_form || (parent && parent.relative_names_in_form?)
-  end
```
ethn: REMOVED API.

we just always use relative names in forms now.  (relative to the form context).  So this stuff is no longer necessary.

cardname/lib/cardname/contextual.rb
```diff
@@ -33,18 +33,21 @@ def from *from
...
-    def context_relevant? remaining
-      remaining.compact.any? && # not all name parts in context
-        remaining != parts      # some name parts in context (could be faster test!)
+    def remove_context *from
+      return false unless from.compact.present?
+      remaining = parts_excluding *from
+      return false if remaining.compact.empty? || # all name parts in context
+                      remaining == parts          # no name parts in context
+                                                  # (could be faster test!)
+      remaining
```
ethn: I don't think this significantly changes the logic of `#name_from`, but it's more readable imo.

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -140,30 +140,53 @@ def process_edit_fields fields
...
+    end
   end
 
-  # If you use subfield cards to render a form for a new card
-  # then the subfield cards should be created on the new card not the existing
-  # card that build the form
-  def name_in_form
-    relative_names_in_form? ? card.relative_name : card.raw_name
+  def form_prefix_from_voo
+    voo&.live_options&.dig :input_name
```
ethn: so...arguably this option should also be renamed to `form_prefix` (or should at least be consistent with other nomenclature).

card/mod/standard/set/all/rich_html/form.rb
xithan: I'm not convinced that @edit_in_form is an improvement. There are a lot of methods with `multi` and `single` in the name to distinguish between those two cases. Since an edit view is always a form, edit_in_form doesn't make it clear to me that it belongs to the multi case.

decko/features/step_definitions/wagn_steps.rb
```diff
@@ -87,7 +87,7 @@
   signed_in_as(username) do
     visit "/card/edit/#{cardname.to_name.url_key}"
     plusses.hashes.first.each do |name, content|
-      set_content "card[subcards][#{cardname}+#{name}][content]", content
+      set_content "card[subcards][+#{name}][content]", content
```
xithan: Why this change? The old way is supposed to work too, no?

card/mod/standard/set/all/rich_html/form.rb
ethn: yeah, I liked the idea of having the view and variable have the same name.

For now I'll just change it so the test and the variable have the same name and use `@in_multi_card_editor`.  (The only other method that mentions this variable is `#in_multi_card_editor?`, which only checks to see whether the variable is present!)  This will satisfy my need to reduce all the naming noise.

decko/features/step_definitions/wagn_steps.rb
ethn: Sorry, I'm wrong.  Yes, the old way would still work in a form.  But in this case, it means "set the content of the field named (X) to (Y).  Since the field now has a relative name, Capybara won't find the field without this change if we use the old name.

PR #16, opened 2017-08-25
Params

ethn:
adapting (clumsily at times) to new rails param handling


decko/rails/controllers/card_controller.rb
```diff
@@ -60,6 +60,13 @@ def setup
     Card::Mod::Loader.refresh_script_and_style if Rails.env.development?
     Card::Cache.renew
     Card::Env.reset controller: self
+    # unprotect_card_params!
```
xithan: not used?

PR #19, opened 2017-08-31
Viewcaching

ethn:
mostly documentation and nomenclature changes (so far)


card/lib/card/view/options.rb
```diff
@@ -176,16 +210,24 @@ def process_default_options
...
       def foreign_live_options
-        foreign_options live_options
+        foreign_options_in live_options
+      end
+
+      # find non-standard option in Hash
+      # @param opts [Hash] options hash
+      # @return [Hash] options Hash
+      def foreign_options_in opts
+        opts.reject { |k, _v| Options.all_keys.include? k }
```
xithan: How do I add foreign options?

card/lib/card/view/options.rb
```diff
@@ -76,38 +94,44 @@ def define_setter option_key
...
 
-      # The following methods comprise the primary voo API.  They allow
-      # developers to read and write options dynamically
 
-      def items
-        live_options[:items] ||= {}
-      end
+      # Developers can also set most options directly via accessors, eg voo.title = "King"
+      #
+      # Note that a few options do not follow this pattern.
```
xithan: Why?

card/lib/card/view/options.rb
ethn:
```
view :myview do |args|
  m = args[:my_foreign_arg]
end

render_myview my_foreign_arg: 4
```
...I'll add a coment to explain.

card/lib/card/view/options.rb
ethn: Thanks.  Forgot to finish that piece.  Good catch.

card/lib/card/view/options.rb
xithan: I thought we want to get rid of args. 
And how do I change the type of the foreign args (like "heir" or "none")?

xithan: Could be worth it to change the view api so that you have to define the foreign options that a view supports and requires.
For example `view :myview, optional: [:optional_arg], mandatory: [:mandatory_arg]` or just `view :myview do |mandatory_arg, optional_arg: nil| do`

ethn: I did want to get rid of args, but the short version is that I couldn't quite yet; there are still too many places where we use them.

Another way to say this is that any thing that comes through in the |args| hash is a foreign option.  If it weren't foreign, it would be part of the voo API. You could justifiably argue that we could just add ways to access the foreign objects; I like your proposal to identify foreign options in the view definition as a way to make that much more compelling.

Really, I think there should be a ViewOption _class_ where we define these things, and the view definition should basically provide a means to define simple options on the fly.

I'll admit that I'm a little nervous to open this can of worms, but it sure would be nice to have a comprehensive solution in place...

ethn: > And how do I change the type of the foreign args (like "heir" or "none")?

So far, that's not possible.  I think that's why you had to invent the `closest_live_option` thing. (foreign options can't be heirs).  Everything that's called on (at least within the gem) is a foreign_option: `followed_set`, `follow_option`, `follower`, and `inline_attachment_url`.

Perhaps some of those wouldn't be needed if they could be heirs?

PR #25, opened 2017-09-06
mods as gems and no tmp files

xithan:


card/mod/admin/spec/set/self/admin_spec.rb
```diff
@@ -38,16 +38,16 @@
...
+        #   Card::Auth.as_bot do
+        #     a = Card['A']
+        #     puts a.references_out.count
+        #     Card::Env.params[:task] = :repair_references
+        #     puts a.references_out.count
+        #     @all.update_attributes({})
+        #     puts a.references_out.count
+        #
+        #   end
+        # end
```
ethn: 🤔

card/mod/core/format/html_format.rb
```diff
@@ -70,6 +70,10 @@ def html_escape_except_quotes s
         s.to_s.gsub(/&/, "&amp;").gsub(/\'/, "&apos;")
          .gsub(/>/, "&gt;").gsub(/</, "&lt;")
       end
+
+      def content_type
```
ethn: any reason you chose this over mime_type?

card/mod/history/set/all/history.rb
```diff
@@ -28,27 +28,31 @@ def actionable?
...
 
 # changes for the create action are stored after the first update
 def store_card_changes_for_create_action
-  changed_fields.each do |f|
+  Card::Change::TRACKED_FIELDS.each do |f|
     Card::Change.create field: f,
-                        value: attribute_was(f),
+                        value: attribute_before_act(f),
                         card_action_id: create_action.id
   end
```
ethn: doesn't this create a Card::Change for fields that weren't changed?

ethn: oh.  Create action.  Doh.

card/spec/models/card/trash_spec.rb
```diff
@@ -200,7 +200,7 @@
     end
   end
 
-  it "does not be trash" do
+  it "isn't trash" do
```
ethn: HAHAHAHAHA

decko/lib/decko/response.rb
```diff
@@ -16,14 +15,14 @@ def card_redirect url
...
         card_redirect result
+      elsif format.is_a?(Card::Format::FileFormat) && status == 200
+        send_file(*result)
       else
-        args = { html: result.html_safe, status: status }
-        args[:content_type] = "text/text" if format == :file
-        render args
+        render body: result.html_safe,
+               status: status,
+               content_type: format.content_type
```
ethn: I'm guessing this is why you used "content_type" rather than "mime_type" as the method, but given how weighted the words "content" and "type" already are for cards, I'm thinking "mime_type" would be less likely to confuse folks (including me)

decko/lib/decko/tasks/decko/bootstrap.rake
```diff
@@ -4,9 +4,16 @@ namespace :decko do
...
+
       delete_unwanted_cards
       Card.empty_trash
+      Rake::Task["decko:bootstrap:copy_mod_files"].invoke
+      [[:all, :script], [:all, :style], [:script_html5shiv_printshiv]].each do |name|
+        Card[*name].make_machine_output_coded
+      end
+      binding.pry
+      clear_history
+      binding.pry
```
ethn: binding.pry x2

decko/lib/decko/tasks/decko/bootstrap.rake
```diff
@@ -107,15 +110,10 @@ def delete_unwanted_cards
...
-  conn.execute("truncate sessions")
+  puts "clearing history"
   act = Card::Act.create! actor_id: Card::WagnBotID,
                           card_id: Card::WagnBotID
-  Card::Action.find_each do |action|
-    action.update_attributes!(card_act_id: act.id)
-  end
+  Card::Action.make_current_state_the_initial_state act
+  #conn.execute("truncate card_acts")
+  ActiveRecord::Base.connection.execute("truncate sessions")
```
ethn: we don't actually use the sessions table any more, right?  so this is harmless but non very meaningful...

card/db/migrate/20170908114442_create_delayed_jobs.rb
```diff
@@ -0,0 +1,22 @@
+class CreateDelayedJobs < ActiveRecord::Migration[5.1]
```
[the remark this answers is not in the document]
xithan: right

card/db/migrate/20170908114442_create_delayed_jobs.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.up
+    create_table :delayed_jobs, force: true do |table|
+      table.integer :priority, default: 0, null: false # Allows some jobs to jump to the front of the queue
+      table.integer :attempts, default: 0, null: false # Provides for retries, but still fail eventually.
+      table.text :handler,                 null: false # YAML-encoded string of the object that will do work
+      table.text :last_error                           # reason for last failure (See Note below)
+      table.datetime :run_at                           # When to run. Could be Time.zone.now for immediately, or sometime in the future.
```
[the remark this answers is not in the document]
xithan: 90

card/lib/card/mod/dirs.rb
```diff
@@ -72,6 +117,19 @@ def load_from_dir
         end.compact
       end
 
+      def load_from_gemfile
```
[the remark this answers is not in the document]
xithan: I think we always have to mix all three options:
gem mods can only be loaded via Gemfiles. Since gems don't have a Gemfile we need the Modfile in the card gem to load the mods of the card gem. If you want to use gem mods in your deck you have to use the Gemfile because otherwise it's not available. Bundler doesn't install mods of the Modfile. For your own local mods you can choose whether you add a Modfile for those in your mod directory or just let decko load everything.

card/lib/card/mod/loader/module_loader/load_strategy.rb
```diff
@@ -0,0 +1,46 @@
...
+            @module_type = module_type # :set or :set_pattern
+            @module_template = module_template
+          end
+
+          private
+
+          def each_file &block
+            @mod_dirs.each @module_type do |base_dir|
+              if @module_type == :set
+                # I'm not sure if we really need this ordering by pattern for sets -pk
```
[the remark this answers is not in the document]
xithan: There is an index option to determine the pattern order. But if that option is not given then it's the reverse load order. But this is not about loading set patterns it's about loading sets

card/lib/card/mod/dirs.rb
```diff
@@ -1,38 +1,83 @@
...
+    # in load order with a preceding "mod" command (similar to a Gemfile).
+    # The mods are expected in subdirectories with the mod names.
+    #
+    # Example for a mod directory:
+    #   # my_mod/Modfile
+    #   mod "twitter"
+    #   mod "cache"
+    #
+    #   # directory structure
+    #   my_mods/
```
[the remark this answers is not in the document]
xithan: It doesn't have to be in your deck. In can also be `card/mod/` or `decko-mods/` or you could add any other directory that looks like that to your mod load paths list.

PR #26, opened 2017-09-08
irreversible admin tasks

xithan:
and fix dirty attributes handling

### DIFF
--- a/card/mod/admin/spec/set/self/admin_spec.rb
+++ b/card/mod/admin/spec/set/self/admin_spec.rb
@@ -9,45 +9,62 @@ describe Card::Set::Self::Admin do
   end
 
   context "#update" do
-    before do
-      @admin = Card[:admin]
-    end
+    let(:admin) {Card[:admin]}
 
-    it "triggers empty trash (with right params)" do
+    def run_admin_task task
       Card::Auth.as_bot do
-        Card["A"].delete!
-        expect(Card.where(trash: true)).not_to be_empty
-        Card::Env.params[:task] = :empty_trash
-        @admin.update_attributes({})
-        expect(Card.where(trash: true)).to be_empty
+        Card::Env.params[:task] = task
+        admin.update_attributes({})
       end
     end
 
-    # NOTE: I removed this functionality for now, because I don't think we
-    # should have web access to admin functions that can incur actual data loss.
+    it "clearing trash is denied" do
+      expect { run_admin_task :empty_trash }
+        .to raise_error Card::Error::Oops, /The admin task 'empty trash'/
+    end
 
-    # it "triggers deleting old revisions (with right params)" do
+    it "clearing history is denied" do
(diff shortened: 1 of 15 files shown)


card/mod/admin/spec/set/self/admin_spec.rb
```diff
@@ -9,45 +9,62 @@
...
         expect(Card.where(trash: true)).not_to be_empty
-        Card::Env.params[:task] = :empty_trash
-        @admin.update_attributes({})
+        run_admin_task :empty_trash
         expect(Card.where(trash: true)).to be_empty
       end
-    end
 
-    # NOTE: I removed this functionality for now, because I don't think we
-    # should have web access to admin functions that can incur actual data loss.
```
xithan: You only removed the method but not the link. I brought the method back but disabled it by default with an error message that explains how to turn it on.

card/mod/history/set/all/content_history.rb
```diff
@@ -166,8 +168,7 @@ def draft_acts
   drafts.created_by(Card::Auth.current_id).map(&:act)
 end
 
-event :detect_conflict, :validate, on: :update,
-                                   when: proc { |c| c.edit_conflict? } do
+event :detect_conflict, :validate, on: :update, when: proc { |c| c.edit_conflict? } do
```
ethn: I think `when: :edit_conflict?` should work.

ethn: (I mean, I think it already works)

decko/lib/decko/config/initializers/sedate_parser.rb
```diff
@@ -0,0 +1,7 @@
+# Hack to get rid of annoying parser warnings
```
ethn: nicely named.

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -264,7 +261,7 @@ def original_filename
     end
 
     def action_id
-      model.selected_content_action_id || NO_ACTION_ID
```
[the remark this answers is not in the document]
xithan: no, it caused problems when I turned off the history for machine output. The generate machine output had no longer a file name.

card/mod/history/set/all/content_history.rb
```diff
@@ -94,25 +98,23 @@ def new_content_action_id
 end
 
 def current_action_changes_content?
-  new_card? || @current_action.new_content? || db_content_changed?
+  new_card? || @current_action.new_content? || db_content_is_changing?
 end
 
 def last_action_id
-  (la = last_action) && la.id
+  last_action&.id
```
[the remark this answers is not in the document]
xithan: That's safe to assume. It's the same as `last_action.id`. The only difference is that it won't fail if last_action is nil. Instead it returns nil.

PR #29, opened 2017-09-14
line correction

xithan:


card/lib/card/format/render.rb
```diff
@@ -100,10 +100,10 @@ def expand_stubs cached_content
         conto.process_each_chunk do |stub_hash|
           yield(stub_hash)
         end
+
         if conto.pieces.size == 1
           # for stubs in json format this converts a single stub back
           # to it's original type (e.g. a hash)
-          # TODO: method name to_s is misleading
           conto.pieces.first.to_s
```
xithan: We can't call `result` because String doesn't have a result method.

PR #34, opened 2017-09-19
@card.codename returns a Symbol (or nil)

ethn:


card/spec/lib/card/set_pattern_spec.rb
```diff
@@ -11,7 +11,7 @@ def it_generates opts
 describe Card::Set::Pattern do
   specify ".in_load_order" do
     expect(Card::Set::Pattern.in_load_order)
-      .to eq(%w[abstract all all_plus type star rstar right type_plus_right self])
+      .to eq(%w[abstract all all_plus type star rstar right type_plus_right self].map &:to_sym)
```
xithan: `%i[...]`

ethn: cool!

PR #35, opened 2017-09-20
Name object

ethn:
An experiment.  not to be merged.


ethn:
Closing.  Moving to inheriting from String

PR #36, opened 2017-09-21
Doc

ethn:
not as much as I wanted, but making progress.

also has a lot of fetch-related cleanup.

### DIFF
--- a/card/lib/card/codename.rb
+++ b/card/lib/card/codename.rb
@@ -3,17 +3,22 @@ require_dependency "card/cache"
 require_dependency "card/name"
 
 class Card
-  # {Card}'s names can be changed, and therefore names should not be directly
-  # mentioned in code, lest a name change break the application. Instead, a
-  # {Card} that needs specific code manipulations should be given a {Codename},
-  # a unique string identifier that will not change even if the card's name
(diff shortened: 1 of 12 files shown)


ethn:
it doesn't render the fetch.rb diff by default, but the biggest change here is all the stuff moved out of there, in part to make it an approachable file to those wanting to learn the API

card/lib/card/codename.rb
```diff
@@ -3,17 +3,22 @@
...
-  # Every process maintains a complete cache that is not frequently reset
+  # An administrator might add to the Company card via the RESTful web API with a url like
+  #
+  #     /update/CARDNAME?card[codename]=CODENAME
+  #
+  # ...or via the api like
+  #
+  #     Card[CARDNAME].update_attributes! codename: CODENAME
+  #
+  # Generally speaking, _codenames_ are represented by Symbols.
```
xithan: Since a decker will never use codenames, it's probably more important to mention how to add codenames with migrations.

ethn: good point.  I've been more focused on making sure people can successfully get started on their own mod rather than best practices for deployment.

This is related to "Make mod installation handle card creation/updates" (Mod Squad doc).  If we did that well, then perhaps the best practice could also be quick/simple enough for a newbie.

PR #38, opened 2017-09-28
Name2

ethn:
Cardname inherits from String.  This made many methods accidental overrides, especially `#+`, `#length`, and `#replace`.  To get things working, I removed or renamed those methods. 

`Name.new` still requires a hash lookup, but initialization does not generate a key.

`#name` now returns a Card::Name object

there is no `#cardname` (we can add back an alias for backwards compatibility if needed).


ethn:
ViewName was causing me a bunch of headaches, and I couldn't understand why it was of any value, so I got rid of it for now.  I can bring it back if there's a good reason to (and someone can explain it to me so I can write tests for it)

xithan:
The value of ViewName was that we could use plural and singular for different views.

ethn:
So, with this PR, views are not treated as names at all (or at least that is the intent), so you can still use plural and singular for different views.  I suppose what you lose is that you can't have your view lower-cased for you (`{{cardname|CLOSED}}`), have punctuation removed, etc.  But do we really need that?

I think I originally started treating views as names, because I thought we wanted singularization.  Now that we don't, is there any good reason left to do any special manipulation of viewnames that might warrant its own class?

xithan:
maybe we want it if we introduce cards as views

ethn:
Yeah, true.  Seems dangerous to have multiple key-generation methods in the same deck.  But I bet we could get around that...

PR #41, opened 2017-09-28
 get rid of active_support and active_record in top level namespace

xithan:


ethn:
even though this is approved already, would you be willing to merge in master so we can see it green before we merge it in?

xithan:
I submit this again to get rid of all the noise

PR #42, opened 2017-09-29
Renderbang

ethn:
as discussed:

> render -> render!
> optional_render -> render


xithan:
`render!` would be also a good name for `_render`

ethn:
That's true.  Fwiw, we did agree on this API change last November:

>did either of you see the API proposal in the #code channel of the decko team?
> short version:
> • render -> render!
> • optional_render -> render

I think we overuse _render, which is mostly a legacy problem – permission checks used to be slow; now they're not (plus view caching will be dramatically reducing our render counts).

I think I would suggest that render! means _both_ non-optional and no permissions check.  And we get rid of the underscore thing.  So we'd be left with this:

```                     
                    check permissions            don't check permissions
_______________________________________________________________________________________
    optional |      render                       render check_permissions: false
not optional |      render optional: false       render!
```

In general, I'd say optimizing performance is a reasonable justification for using `render!` in tests but not a great justification for using it live; every time we skip permissions it's a potential security weakness. 

If necessary we should optimize permission and option checking.

If you agree with this, we can go ahead and approve this PR as a step in the right direction.

PR #46, opened 2017-10-04
fetch fix

xithan:


card/spec/support/matchers/card_matchers.rb
```diff
@@ -40,21 +40,45 @@ def expect_card *marks
...
+
+  failure_message do |card|
+    return super(card) unless @field
+    if @content
+      "expected #{card} to have a field '#{field_key}' with content '#{@content}',
+but content is #{card.content.present? ? "empty" : card.content}"
+    elsif @pointing_to
+      "expected #{card} to have a field #{field_key} pointing to #{@pointing_to} but
+content is #{card.content}"
+    end
```
ethn: I think this is missing an `end`, which is why rspecs are failing.

card/mod/history/set/all/history.rb
```diff
@@ -83,7 +83,7 @@ def finalize_action?
 event :remove_empty_act,
       :integrate_with_delay_final,
       when: proc { |c|  c.act_card? } do
-  if @current_act&.actions&.reload.empty?
+  if @current_act&.actions&.reload&.empty?
```
ethn: generally speaking, I think these kinds of things should be in a :when test.

(I know you didn't add it here)

The question is whether something an `after: :remove_empty_act` event should fire, even if no empty act was removed.  If this is in `when:`, it doesn't fire.

In fact, we might even start a pattern where if I make `def remove_empty_act?`, then this is automatically treated as the test?  hmm...

card/mod/core/set/all/fetch_helper.rb
```diff
@@ -98,8 +98,8 @@ def new_card_fetch_results card, mark, opts
       when opts[:skip_virtual] then return nil
     end
     card.assign_name_from_fetched_mark! mark, opts
-    return nil unless opts[:new] || card.known?
     finalize_fetch_results card, opts
+    card if opts[:new] || card.known?
```
ethn: sorry

card/spec/support/matchers/card_matchers.rb
xithan: no, two `end`s :simple_smile:

card/mod/history/set/all/history.rb
xithan: Sounded like a good idea but becomes worse the more you think about it. The only win is that we save a few key strokes but if the check method isn't close to the event the logic is obscured and it will become a nightmare of debugging if you create a check method that accidentally matches an existing event or vice versa.

ethn: Yeah, I agree.  maybe just a convention (specified explicitly)

PR #49, opened 2017-10-09
made seed_consts handling work in normal gem context

ethn:


decko/lib/decko/tasks/decko.rake
```diff
@@ -1,6 +1,6 @@
 require "decko/application"
 require_relative "alias"
-require_relative "../../../../card/db/seed_consts"
```
ethn: card isn't just "card" when installed as a normal gem.

I moved seed_consts into lib as the most obvious way to make it available. Open to better ideas.

PR #50, opened 2017-10-09
Sourcecache

ethn:


card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -168,12 +172,13 @@ def filename
     end
 
     def extension
-      case
-      when file && file.extension.present? then ".#{file.extension}"
-      when card_content = model.content    then File.extname(card_content)
-      when orig = original_filename        then File.extname(orig)
-      else                                   ""
-      end.downcase
+      @extension ||=
```
xithan: That looks dangerous to me. What if the file changes?

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -159,6 +159,10 @@ class FileCardUploader < Uploader::Base
     ].freeze
     delegate :store_dir, :retrieve_dir, :file_dir, :mod, :bucket, to: :model
 
+    def valid?
+      extension.present?
```
xithan: what about files without extension?

card/mod/carrierwave/set/type/image.rb
```diff
@@ -98,17 +99,18 @@ def new_image action
...
+      return source unless url_generator && ::File.exist?(path)
+      image_tag url_generator.call(path)
+    end
   end
 end
 
 format :css do
   view :core do
-    render_source
+    handle#_source
```
xithan: ?!

card/mod/carrierwave/set/type/image.rb
```diff
@@ -119,16 +121,16 @@ def new_image action
...
-  def selected_file_version
-    style = voo.size = image_style.to_sym
-    if style && style != :original
-      card.attachment.versions[style]
-    else
-      card.attachment
-    end
-  end
+  #def image_style
+  #  ["", "full"].member?(params[:size].to_s) ? :original : params[:size].to_sym
```
xithan: We drop support for `full` and a size parameter?

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
ethn: I couldn't see how these objects could persist from one request to the next, so I didn't think there was much risk here.  However, the reward may not be big enough for much risk at atll, although I did see extension getting called a few different times and wondered whether we were repeating that processing.

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
ethn: Currently we can't handle those.  We generate a url that ends in a period (eg `files/~12/345.`), and then decko misinterprets the url as requesting the file in html format.  This translates into fairly expensive requests.

I'll add a note that says it should eventually be valid for files not to have an extension.  But right now I think the only way you'd actually be able to access such a file is (maybe?) with the `.file` extension, which is weird decko functionality almost nobody knows about.

card/mod/carrierwave/set/type/image.rb
ethn: oops.

card/mod/carrierwave/set/type/image.rb
ethn: Still supported!  The old (current) implementation has both `selected_version` and `selected_file_version` doing the same thing.  See file.rb, where the `selected_file_version` is replaced with `selected_version` (which file and image handle differently)

card/mod/core/set/all/fetch_helper.rb
```diff
@@ -72,13 +72,16 @@ def retrievable_mark_type_and_value mark
     end
   end
 
-  def retrieve_from_db? card, opts
-    card.nil? || (opts[:look_in_trash] && card.new_card? && !card.trash)
+  def retrieve_from_db? card, mark_type, mark_key, look_in_trash
+    return false if card&.real?                                      # real card found in cache
+    return false if mark_type === :key && mark_key.to_name.relative? # no relative names in db
+    return true if card.nil?                                         # card not found in cache
+    look_in_trash && !card.trash
```
ethn: I should flag that this is another change in fetch logic.  I was seeing lots of lookups in the logs for things like "+topic", which we know is not going to be in the database, so I added the line about "no relative names".

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
xithan: not request but what about file updates?

xithan:
```ruby
card.attachment.extension # => ".csv"
card.update_attributes! file: File.open("/Users/philipp/log.txt")
card.attachment.extension # => ".csv"
```

xithan: without your caching the code above returns the correct updated extension

card/mod/core/set/all/fetch_helper.rb
xithan: We don't allow `+topic` but it's possible to create cards like `_main`, `_user`, `_left` and `relative?` is true for those,

card/mod/core/set/all/fetch_helper.rb
```diff
@@ -72,13 +72,16 @@ def retrievable_mark_type_and_value mark
     end
   end
 
-  def retrieve_from_db? card, opts
-    card.nil? || (opts[:look_in_trash] && card.new_card? && !card.trash)
+  def retrieve_from_db? card, mark_type, mark_key, look_in_trash
+    return false if card&.real?                                      # real card found in cache
+    return false if mark_type === :key && mark_key.to_name.relative? # no relative names in db
+    return true if card.nil?                                         # card not found in cache
```
xithan: a minor style issue: I think `card.nil?` is the main case that this method  covers but with the return statement it became more like a exceptional  state.

card/mod/core/set/all/fetch_helper.rb
ethn: should that be allowed?

Really the problem for me was that it kept looking in the db for the relative cards load after load.  So they weren't getting in the cache AND they kept making db calls.

xithan: Probably better not to allow it. Even if we make them fetchable again they will have other issue (in nests for example). 

Do you know where these fetch calls for relative cards come from? Is that also bad code or justified for some reason?

ethn: I'm not sure where it comes from, but my guess is that it's broadly used in places where we don't know what kind of name we're going to get.  Eg, we might look up a nested card with something like `Card.fetch nestedcardname, new: {}`, and in the case of a new form, all the field names will look like this.

I think it's fine to support fetching relative cards and changing their name later.  You can still do `Card.fetch("+topic", new: {})` and get a response, but we won't bother looking in the database for it.  (I wrote it to get silly queries out of the log).

Fwiw, I don't think this changes any behaviors; it's just an optimization. I only wanted to flag it because the last time I changed things around in fetch_helpers, I accidentally broke stuff.  

But are we agreeing that we don't want to allow cards with relative names to get stored in the database?  Basically, we just need to validate that a name is not relative when we do all our other name validations, right?

card/lib/card/view/stub.rb
```diff
@@ -11,7 +11,7 @@ class View
     module Stub
       # @return [String]
       def stub
-        "<card-view>#{stub_json}</card-view>".html_safe
+        "(stub)#{stub_json}(/stub)".html_safe
```
ethn: FYI, this is our new stub syntax.  We should probably discuss.

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
ethn: I ended up not using this check for now, because it was killing a bunch of wikirate specs where we didn't have valid images (logo, Samsung+image, missing, etc).  The check is still here but not used in source view anymore, so for now it does nothing.

ethn: hmm correction: I guess it's still used on File cards but not Images

card/mod/carrierwave/set/type/image.rb
```diff
@@ -9,46 +9,60 @@
...
   def determine_image_size
     voo.size =
       case
-      when nest_mode == :closed then :icon
+      when nest_mode == :closed then closed_size
       when voo.size             then voo.size.to_sym
-      when main?                then :large
-      else                           :medium
+      when main?                then main_size
+      else                           default_size
```
ethn: this is a newer change and means different formats can have different default sizes.  Potentially useful..

card/mod/core/chunk/view_stub.rb
```diff
@@ -4,9 +4,9 @@ module Chunk
       class ViewStub < Abstract
         Chunk.register_class(
           self,
-          prefix_re: Regexp.escape("<card-view>"),
-          full_re: /\<card-view\>([^\<]*)\<\/card-view\>/,
-          idx_char: "<"
+          prefix_re: Regexp.escape("(stub)"),
+          full_re: /\(stub\)([^\(]*)\(\/stub\)/,
+          idx_char: "("
         )
```
ethn: In case you, like me, have forgotten how this works, each Chunk has a different "idx" character.  Before this was "<", but now it's "(".  Because of the `[^\(]` in the regexp, I think I need to add parenthesis escaping to the stub generator.

Or maybe there's a better regexp...

card/mod/core/set/all/fetch_helper.rb
```diff
@@ -72,13 +72,15 @@ def retrievable_mark_type_and_value mark
     end
   end
 
-  def retrieve_from_db? card, opts
-    card.nil? || (opts[:look_in_trash] && card.new_card? && !card.trash)
+  def retrieve_from_db? card, mark_type, mark_key, look_in_trash
+    return false if card&.real?                                      # real card found in cache
+    return false if mark_type == :key && mark_key.to_name.relative?  # no relative names in db
+    card.nil? || (look_in_trash && !card.trash)                      # card not found in cache (or trash lookup)
```
ethn: better?

ethn:
unless you object, I'm probably going to merge this into master, since (a) it's green and (b) it's merged into wikirate already.

card/lib/card/view/stub.rb
xithan: I think that violates html specs. Custom tags are supposed to have a "-" in the name.

xithan: oh, it's not even a tag anymore.

xithan: I guess if I write "(stub)" in a card everything goes to hell. Isn't it better to change the tag strip algorithm so that it doesn't remove custom tags?

ethn: let's discuss.  the strip tag solution was my first impulse until I saw that there were lots of places where there was stripping going on using several different mechanisms, and I didn't want to update them all.  maybe I'll change it to something slightly less likely to be stumbled upon

PR #54, opened 2017-10-14
changes for wikirate

xithan:


card/lib/card/format/render.rb
```diff
@@ -73,11 +73,13 @@ def view_cache_setting view
       end
 
       def stub_render cached_content
-        expand_stubs cached_content do |stub_hash|
+        result = expand_stubs cached_content do |stub_hash|
           prepare_stub_nest(stub_hash) do |stub_card, mode, options, override|
             with_nest_mode(mode) { nest stub_card, options, override }
           end
         end
+        puts "STUB IN RENDERED VIEW: #{result}" if result =~ /stub/
```
ethn: I see my code made it in :)

I want to make this a bit noisier and leave this around for a bit until we stop seeing stubs where we don't want them.

card/mod/account/set/right/password.rb
```diff
@@ -21,7 +21,7 @@
       on: :save, changed: :content,
       when: proc { !Card::Env[:no_password_encryptions] } do
   # no_password_encryptions = hack for import - fix with api for ignoring events
-  salt = get_salt
+  salt = left && left.salt
```
ethn: this one, on the other hand, was supposed to be temporary...

PR #61, opened 2017-10-20
Storage type

ethn:


card/mod/core/set/all/actify.rb
```diff
@@ -31,25 +32,16 @@ def start_new_act opts
   ActManager.clear
   self.director = nil
   ActManager.act_card = self
-  if opts && opts[:success]
-    Env[:success] = Env::Success.new(name, Env.params[:success])
-  end
-  run_callbacks :act do
-    yield
-  end
+  Env.success(name) if opts[:success]
```
xithan: The not nil check was probably there for reason (just because the default is a hash doesn't mean that it can't be nil). If you remove it make at least sure that it is a hash. An `opts ||= {}` at the beginning of the method will probably save us a ton of airbrake messages and a debugging chore.

decko/features/support/env.rb
```diff
@@ -20,7 +20,7 @@
 # newer version of cucumber-rails. Consider adding your own code to a new file
 # instead of editing this one. Cucumber will automatically load all
 # features/**/*.rb files.
-Before("@background-jobs or @delayed-jobs or @javascript") do |scenario|
+Before("@delayed-jobs or @javascript") do |scenario|
```
xithan: Have you thought this through, changing the database cleaner strategy for @background-jobs? If you wanted to remove it then you should remove it completely or, what I prefer in this case, comment the rest of background-jobs out an leave a comment about this change in case it will be needed again. I don't like an unused api to be crippled.

PR #62, opened 2017-10-20
better view canonicalization

ethn:


card/lib/card/view.rb
```diff
@@ -18,9 +18,16 @@ class View
...
-      return if view.blank? # error?
-      view.to_sym
-#      view.to_viewname.key.to_sym
+      case view
+      when blank? # error?
+        nil
+      when Symbol
+        view
+      when String
+        view.gsub(/\W/, "").to_sym
```
ethn: you could make the case that Viewname was better than this because it cached this gsub.  But it was a whole lot of code for that!  We can probably do a little per-process cache here.

xithan: Instead of this we could fix the bad view call in wikirate. We don't have to correct views if somebody puts wrong characters in it. Just a "view doesn't exist" is also fine.

xithan:
One spec needs to be adjusted to the corrected quotes

PR #66, opened 2017-10-24
salvaged changes from old name_object branch

ethn:
nothing important.  just trying to clean up old branches.

### DIFF
--- a/card/lib/card/subcards.rb
+++ b/card/lib/card/subcards.rb
@@ -206,14 +206,8 @@ class Card
     end
 
     def prepend_plus name
-      case name
-      when Symbol
-        "+#{Card[name].name}"
-      when /^\+/
-        name
-      else
-        "+#{name}"
-      end
+      name = name.to_name
+      name =~ /^\+/ ? name : "+#{name}"
     end
 
     def field_name_to_key name
--- a/card/lib/card/subcards.rb
+++ b/card/lib/card/subcards.rb
@@ -54,7 +54,8 @@ class Card
     end
 
     def add_child name, args
-      add prepend_plus(name), args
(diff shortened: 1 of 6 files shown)


card/lib/card/subcards.rb
```diff
@@ -206,14 +206,8 @@ def fetch_subcard key
...
-      case name
-      when Symbol
-        "+#{Card[name].name}"
-      when /^\+/
-        name
-      else
-        "+#{name}"
-      end
+      name = name.to_name
+      name =~ /^\+/ ? name : "+#{name}"
```
xithan: The method treated symbols as codenames before this change, now it just converts them to names.

xithan:
16 failing specs agree with my judgement :wink:

cardname/lib/cardname/manipulate.rb
```diff
@@ -36,6 +36,11 @@ def num_parts
       parts.length
     end
 
+    def prepend_joint
+      joint = self.class.joint
+      self =~ /^#{Regexp.escape joint}/ ? self : (joint + self)
+    end
```
ethn: Henry actually pinged me recently to ask about using different joints.  It would probably be good to look for easy opportunities to change `+` to `Card::Name.joint`.

card/mod/standard/set/type/set.rb
```diff
@@ -302,15 +302,12 @@ def follow_label
...
-    end
-  else
-    "#{name}+#{Card[:all].name}+#{Card[:follow].name}"
-  end
+  follower = case user
+             when nil    then :all.cardname
+             when String then user
+             else             user.name
+             end
+  [name, follower, :follow.cardname].join "+"
```
xithan: You missed to replace a "+".

PR #71, opened 2017-10-26
Handle missing codenames in Card#fetch_name

xithan:


card/mod/core/set/all/fetch.rb
```diff
@@ -88,6 +88,8 @@ def fetch_name *mark
     elsif block_given?
       yield
     end
+  rescue Card::Error::NotFound => e
```
ethn: suggested comment here:
> # eg. if codename is missing

also, just noticing that there is a documentation mismatch.  to be consistent, we should probably use `*args` instead of `*mark` and should send `*args` (with `*`) to quick_fetch, no?

xithan: The way `quick_fetch` is implemented it should be `*mark` in quick_fetch since it doesn't support additional args.

ethn: ok, I can fix those.

PR #72, opened 2017-10-26
haml template path bug patch

ethn:


card/mod/core/chunk/view_stub.rb
```diff
@@ -48,6 +48,10 @@ def interpret_override
         def process_chunk
           @processed = yield @stub_hash
         end
+
+        def result
+          @processed
+        end
```
ethn: not sure how this got here, but I remember that this patch was to make it so that views could return nil without confusing the view cache.

card/mod/search/set/all/filter.rb
```diff
@@ -5,8 +5,8 @@ def filter_form data={}, form_args={}
     render_haml :filter_form, categories: data, form_args: form_args
   end
 
-  def view_template_path view
-    super(view, __FILE__)
+  def view_template_path view, set_path= __FILE__
+    super(view, set_path)
```
ethn: This is in an `all` set, so overriding the other `view_template_path` without a path argument breaks a bunch of thins in wikirate.

I'm not loving this API, because there's no easy way for other `all` sets to use haml templates without overriding each other.  I think we need to look into the haml API, perhaps in keeping with your `@todo` comments in haml.rb.  More pair coding!

PR #73, opened 2017-10-27
Fetch name

ethn:


card/mod/core/set/all/fetch.rb
```diff
@@ -88,7 +88,7 @@ def fetch_name *mark
     if (card = quick_fetch(mark))
       card.name
     elsif block_given?
-      yield.name
+      yield
```
xithan: This needs a `to_name` conversion so that you don't get the wrong type if the block returns a string (happens in wikirate). It's also wrong in the rescue block

card/lib/card/codename.rb
```diff
@@ -29,7 +29,7 @@ def [] codename
         when Integer
           codehash[codename]
         when Symbol, String
-          codehash.key?(codename.to_sym) && codename.to_sym
+          codehash.key?(codename.to_sym) ? codename.to_sym : nil
```
xithan: `id` uses the same code. Should be corrected there too

PR #75, opened 2017-10-27
Instantiate

ethn:
Old stuff from wagn/wagn made new.


card/lib/card/query.rb
```diff
@@ -129,38 +129,84 @@ def default_comment
...
+    def simple_result? retrn
+      retrn == "count"
+    end
+
+    def name_result record, pattern
+      process_name record["name"], pattern
+    end
+
+    def process_name name, pattern
+      name = pattern.to_name.absolute(name) if pattern =~ /_\w+/
```
ethn: the biggest conflict to resolve since I last worked on this was with the name contextualization, especially the "contextual name processor".  It basically seems to have done the same thing as `pattern.to_name.absolute(name)`, however (as you can see in the spec), it would return a card with the name `""`, whereas the behavior of `#absolute` (for better or for worse) is just to return the context if the name doesn't match the context. 

For now I changed the spec, but I'm not sure what we really want.  Are we ok with queries returning cards with no name?

In either case, I think we probably want that processed in Card::Name as it now is.  But we might want to reconsider (or add more options to) the absolutization behavior

card/lib/card/query/sql_statement.rb
```diff
@@ -49,7 +49,7 @@ def fields
       def full_field table, field
         case field
         when :raw      then "#{table}.*"
-        when :card     then "#{table}.name"
+        when :card     then "#{table}.*"
```
ethn: This is, in some ways, the biggest change.  Instead of having the sql just return a name, it returns all the card info.  In cases where the cache is not populated, this means we can return 20 cards in 1 request as opposed to 21 (our current normal).   But, when the cache is populated, there is extra info getting passed from the database; I don't know how much of a cost that is.  But my notion is that as we hone this, we could make a smart guess about whether or not we're likely to need the extra fields by sticking a query key in the cache.  IF the query has expired, most likely some of the results have, too.

PR #76, opened 2017-10-30
Fix import task

xithan:
Changing the return type of `Card#name` and `#type` from String to
Card::Name broke it.


xithan:
The issue was that `to_yaml` works differently. I don't think we use that somewhere else.

PR #77, opened 2017-10-30
Filter interface

xithan:
and fix render bug


card/mod/core/set/all/fetch.rb
```diff
@@ -90,8 +90,8 @@ def fetch_name *mark
     elsif block_given?
       yield.to_name
     end
-  rescue Card::Error::NotFound => e  # eg. if codename is missing
-    block_given? ? yield.name : raise(e)
+  rescue Card::Error::UnknownCodename => e  # eg. if codename is missing
```
ethn: nice.  probably don't need the comment now ;)

card/lib/card/format/method_delegation.rb
```diff
@@ -0,0 +1,44 @@
+class Card
+  class Format
+    module MethodDelegation
```
ethn: I like this break out.

PR #78, opened 2017-10-31
Rawcontent

ethn:
moving `view: raw` to `def content`


card/mod/account/set/right/salt.rb
```diff
@@ -1,5 +1,5 @@
 include All::Permissions::Accounts
 
-view :raw do |_args|
+view :raw do
```
ethn: this is a case where raw makes sense, imo.  The raw view is effectively an override of the card content (which needs to be kept private in this case)

card/mod/account/set/right/account.rb
```diff
@@ -60,15 +60,11 @@ def token_path_opts
...
-    #     field_nest(:password, view: :titled, title: "password")
-    #   ]
-    # )
-    %({{+#{Card[:email].name}|titled;title:email}}
-      {{+#{Card[:password].name}|titled;title:password}})
+    # Problem: when you do that then the fields are missing in the sign up form:
+    # output( [field_nest(:email, view: :titled, title: "email"),
+    #          field_nest(:password, view: :titled, title: "password")])
+    %({{+#{:email.cardname}|titled;title:email}}
+      {{+#{:password.cardname}|titled;title:password}})
```
ethn: I'm not sure why `def content` didn't work here.  Future research project.

card/mod/account/set/right/email.rb
```diff
@@ -1,17 +1,5 @@
 include All::Permissions::Accounts
 
-view :raw do
-  case
-  when card.real? then card.content
-  # following supports legacy behavior
-  # (should be moved to User+*email+*type plus right)
```
ethn: I did as this comment suggested.

card/mod/admin/set/self/version.rb
```diff
@@ -2,8 +2,8 @@
 
 require_dependency "card/version"
 
-view :raw do |_args|
+def content
   Card::Version.release
 end
 
-view :core, :raw
+# view :core, :raw
```
ethn: These aliases break without a raw view defined in this module.

I want to suggest that if a "hard alias" is impossible, we should consider a "soft" alias (which just generates something like:
```
  view :core do
    render_raw! :raw
  end
```
The use in this case (and several other cases) is just that it avoids having to run "process_content", because we know there's nothing to parse.  Might be better ways...

ethn: eg `chunk_types nil` at the top of the module...

card/mod/account/set/right/password.rb
```diff
@@ -42,6 +30,22 @@
...
+end
+
+format :html do
+  view :core do
+    "<em>#{render_raw}</em"
+  end
+
+  view :editor do
+    card.content = ""
+    password_field :content, class: "d0-card-content", autocomplete: autocomplete?
```
ethn: I wish we didn't have to write "d0-card-content" all the time...

PR #79, opened 2017-11-02
rework of bootstrap.rake

ethn:
not ready to merge: new `rake decko:bootstrap:update` is failing on scss issues

### DIFF
--- a/decko/lib/decko/config/environments/development.rb
+++ b/decko/lib/decko/config/environments/development.rb
@@ -80,7 +80,7 @@ Decko.application.class.configure do
 
   config.action_mailer.perform_deliveries = false
 
-
+  config.active_job.queue_adapter = :inline
 
   # Use Pry instead of IRB
   silence_warnings do
--- a/decko/lib/decko/config/environments/development.rb
+++ b/decko/lib/decko/config/environments/development.rb
@@ -80,7 +80,7 @@ Decko.application.class.configure do
 
   config.action_mailer.perform_deliveries = false
 
-  config.active_job.queue_adapter = :inline
+  config.active_job.queue_adapter = :delayed_job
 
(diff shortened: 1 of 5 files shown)


decko/lib/decko/config/environments/development.rb
```diff
@@ -80,7 +80,7 @@
 
   config.action_mailer.perform_deliveries = false
 
-
+  config.active_job.queue_adapter = :inline
```
ethn: default was `:async`.  that's not a good mix with `allow_concurrency = false`, especially in rake tasks.

decko/lib/decko/tasks/decko/bootstrap.rake
```diff
@@ -1,124 +1,146 @@
 namespace :decko do
   namespace :bootstrap do
-    desc "rid template of unneeded cards, acts, actions, changes, " \
-         "and references"
+    desc "reseed, re-clean, and re-dump"
+    task update: :environment do
+      %w(reseed bootstrap:clean bootstrap:dump).each do |task|
+        Rake::Task["decko:#{task}"].invoke
+      end
+    end
```
ethn: I started working on this because I wanted to get new decks running without having to update machine output.  got into a bit further than intended.

decko/lib/decko/config/environments/development.rb
xithan: This kills our migrations because after a first card change in a migration the whole card namespace is gone including `Card::Auth.current_id`. It seems to work with `config.active_job.queue_adapter = :delayed_job`.

ethn: wow.  any idea why :inline would make us lose the card namespace??

PR #84, opened 2017-11-03
Bootswatch update

xithan:


card/mod/bootstrap/set/abstract/bootstrap_code_file.rb
```diff
@@ -1,42 +1,45 @@
+
 def self.included host_class
   host_class.include_set Abstract::CodeFile
+  host_class.include OverrideCodeFile
   host_class.mattr_accessor :stylesheets_dir, :bootstrap_path, :mod_path
   host_class.mod_path = File.join Cardio.gem_root, "mod", host_class.file_content_mod_name
   host_class.stylesheets_dir = File.join host_class.mod_path, "lib", "stylesheets"
   host_class.bootstrap_path = File.join host_class.mod_path, "vendor", "bootstrap", "scss"
 end
```
ethn: at some point we may want to abstract the parts of this that are just "CodeMultiFile".   Not that much of this code is really bootstrap specific.

decko/lib/decko/generators/decko/templates/Gemfile
```diff
@@ -1,18 +1,21 @@
...
 
 <% if @gem_path.present? %>
-gem 'decko',    :path=>"<%= @gemfile_gem_path %>/decko"
-gem 'card',     :path=>"<%= @gemfile_gem_path %>/card", :require => false
-gem 'cardname', :path=>"<%= @gemfile_gem_path %>/cardname", :require => false
+path "<%= @gemfile_gem_path %>" do
+  gem 'decko'
+  gem 'card', :require => false
+  gem 'cardname', :require => false
+end
```
ethn: oh, I like this.

decko/lib/decko/generators/decko/templates/Gemfile
```diff
@@ -1,18 +1,21 @@
...
 <% else %>
 gem 'decko'
 <% end %>
 
 gem '<%= database_gemfile_entry.name %>'<%= %(, '#{database_gemfile_entry.version}') if database_gemfile_entry.version %>
 
-# Use Unicorn or Thin as server
+# Use Thin or Unicorn as server
```
ethn: if server not specified, it will use webrick, yes?

I should probably add some copy here to refer to our web docs on hosting in production

card/mod/bootstrap/set/abstract/bootstrap_code_file.rb
xithan: CodeFile already handles multiple files (all you have to do is return an array of filenames in `source_files`).
All the code here is about special paths to (bootstrap) stylesheet directories.

decko/lib/decko/generators/decko/templates/Gemfile
xithan: yes. webrick gives a lot of `content-length` warning messages when it serves files. thin doesn't care about that. That's why I switched.

PR #95, opened 2017-11-09
Closed content

srivig:
Fixed: 
-Closed content #152664771
-Pagination #152648884

### DIFF
--- a/card/mod/pointer/set/abstract/01_paging.rb
+++ b/card/mod/pointer/set/abstract/01_paging.rb
@@ -51,7 +51,7 @@ format :html do
   end
 
   def page_link text, page, options
-    return content_tag(:div, text.html_safe, class:"page-link") unless page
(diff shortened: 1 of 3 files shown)


card/mod/pointer/set/abstract/01_paging.rb
```diff
@@ -45,14 +45,14 @@ def paging_links
...
+    wrap_with :li, class: "page-item #{status}" do
       page_link text, page, options
     end
   end
 
   def page_link text, page, options
-    return text unless page
-    options.merge! class: "card-paging-link slotter",
+    return content_tag(:div, text.html_safe, class: "page-link") unless page
+    options.merge! class: "card-paging-link slotter page-link",
```
ethn: it might be nice to change "page" to "paging", because that's less ambiguous (we use "page" to mean other things)

PR #104, opened 2017-11-10
Cache

ethn:
- moving inline Rails.env references to Cardio.config vars
- move cache methods into cache set module
- "cautious" machine refresh checks for file changes after cache is cleared.


card/mod/core/spec/set/all/permissions_spec.rb
```diff
@@ -99,7 +99,12 @@ def assert_not_locked card, msg=""
     it "reader setting" do
       Card.where(trash: false).each do |ca|
         rule_id, rule_class = ca.permission_rule_id_and_class(:read)
-        next if rule_class != ca.read_rule_class
+        if rule_id != ca.read_rule_id
+          puts "read rule id mismatch for #{ca.name} " \
+               "expected #{rule_class}/#{rule_id}, " \
+               "got #{ca.read_rule_class}/#{ca.read_rule_id}"
```
xithan: This is already part of the expect statements. To change the failure message pass the string as second argument to `to`.

card/mod/standard/set/all/rich_html/toolbar.rb
```diff
@@ -143,13 +143,12 @@ def nests_editable?
...
     toolbar_split_button "account", related: :account, icon: :account_box do
-      {
-        account: link_to_related(:account, "details", path: { view: :edit }),
-        roles:   link_to_related(:roles,   "roles"),
-        created: link_to_related(:created, "created"),
-        edited:  link_to_related(:edited,  "edited"),
-        follow:  link_to_related(:follow,  "follow")
-      }
+      %i[ account roles created edited follow ].each_with_object({}) do |item, hash|
+        label = item == :account ? "details" : item.to_s # TODO: i18n
```
xithan: If you do the todo then you don't need the ? operator.

card/config/initializers/01_core_extensions/object.rb
```diff
@@ -3,6 +3,8 @@ module Object
     def deep_clone
       case self
       when Fixnum, Bignum, Float, NilClass, FalseClass, TrueClass, Symbol
+        # FIXME: Fixnum and Bignum are deprecated in Ruby 2.4 but need to be supported here
+        # so long as decko supports older ruby versions
```
xithan: Do we need `Integer` here?

PR #105, opened 2017-11-13
make rich_text editor configurable

ethn:
(made PR to wikirate so changes would be more visible)

### DIFF
--- a/card/lib/cardio.rb
+++ b/card/lib/cardio.rb
@@ -60,6 +60,7 @@ module Cardio
         file_storage:           :local,
         file_buckets:           {},
         file_default_bucket:    nil,
+        rich_text_editor:       :prosemirror,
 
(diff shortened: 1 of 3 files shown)


card/lib/cardio.rb
```diff
@@ -60,6 +60,7 @@ def default_configs
         file_storage:           :local,
         file_buckets:           {},
         file_default_bucket:    nil,
+        rich_text_editor:       :prosemirror,
```
xithan: I think we should return to tinymce as default editor for now. The current version has improved a lot compared to what we were using before we switched to prosemirror. Prosemirror is probably more powerful in the long term (eg. regarding simultaneous editing) but without any plugins the ui is very limited. There are almost no plugins available yet and we don't have the resources to help to change that. PR with tinymce update is coming.

PR #108, opened 2017-11-16
Autocomplete

ethn:
pointer reorganization, new (not great) haml template finding, and autoCardplete concept..


card/mod/core/set/all/haml.rb
```diff
@@ -41,53 +41,43 @@
...
+  rescue => e
+    raise Card::Error, "HAML error #{template_name}: #{e.message}"
+  end
 
-  def view_template_path view, tmp_set_path=__FILE__
-    set_path =
-      tmp_set_path.gsub(%r{/tmp/set/mod\d+-([^/]+)/}, '/mod/\1/template/')
-    haml_template_path view, set_path
+  def identify_template_path view, locals={}
+    deleteme = caller_locations
```
xithan: ?

card/mod/core/set/all/haml.rb
```diff
@@ -41,53 +41,43 @@
...
+    base_path = locals.delete(:template_path) || caller_locations[2].path
+    haml_template_path view, base_path
+  rescue => e
+    binding.pry
```
xithan: !

card/lib/card/set/format/haml_views.rb
```diff
@@ -84,6 +80,14 @@ def template_location debug_info
           Pathname.new(debug_info[:path])
                   .relative_path_from(Pathname.new(Dir.pwd))
         end
+
+        def with_template_path path
+          old_path = @template_path
+          @template_path = path
+          result = yield
+          @template_path = old_path
```
xithan: use `ensure`

card/mod/pointer/set/abstract/02_pointer/options.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def pointer_option_description option
+    pod_name = card.rule(:options_label) || "description"
+    dcard = Card["#{option}+#{pod_name}"]
+    return unless dcard && dcard.ok?(:read)
+    with_nest_mode :normal do
+      subformat(dcard).render_core
+    end
+  end
+end
```
xithan: eof

card/mod/pointer/template/abstract/02_pointer/edit/list_item.haml
```diff
@@ -0,0 +1,9 @@
+%li.pointer-li.mb-1
+  %span.input-group
+    %span.input-group-addon.handle
+      = icon_tag :reorder
+    = text_field_tag "pointer_item", @item, class: "pointer-item-text form-control",
+                                             "data-options-card": @options_card
+    %span.input-group-btn
+      %button.pointer-item-delete.btn.btn-secondary{:type => "button"}
+        = icon_tag :remove
```
xithan: eof

card/mod/account/set/type/signup.rb
```diff
@@ -42,14 +42,9 @@ def account_formgroups
     end
   end
 
-  def view_template_path view, set_path= __FILE__
-    super(view, set_path)
-  end
```
xithan: 👍

card/lib/card/set/format/haml_views.rb
```diff
@@ -12,57 +13,52 @@ module Format
...
-          template = ::File.read template_path
+        def haml_template_proc template, path, &block
+          block_locals = block_given?
           proc do |view_args|
-            voo = View.new(self, view, view_args, @voo)
-            with_voo voo do
-              haml_to_html template, view_args, nil, path: template_path
+            with_template_path path do
+              locals = block_locals ? haml_block_locals(view_args, &block) : view_args
+              haml_to_html template, locals, nil, path: path
```
ethn: I should point out that I removed all three `with_voo` calls from the haml processing in this refactor. They seemed incorrect to me, because the haml blocks should only be called by final_render, which is already within the right with_voo block, so these with_voos were adding a superfluous voo object.

But I often get into trouble by removing things when I don't understand their purpose, so I should get your blessing on this one ;)

card/mod/admin/set/self/admin_info.rb
```diff
@@ -15,17 +15,17 @@ def clean_html?
     end
     return "" if warnings.empty?
     alert :warning, true do
-      render_warning_list warnings
+      warning_list warnings
     end
   end
 
-  def render_warning_list warnings
+  def warning_list warnings
```
ethn: I think we should avoid format methods beginning with `render` that don't render views.  Eg, if someone tried to make a view named "warning list", it could be very confusing.

card/mod/basic_formats/set/all/json.rb
```diff
@@ -38,11 +39,7 @@ def show view, args
   end
 
   def name_search query_attribute: :complete, limit: AUTOCOMPLETE_LIMIT
-    # context is "" so that term will not be interpreted in the context
-    # of search card name.  However, this can break searches where the
-    # search card name is required (eg found_by)
```
ethn: I removed `context: ""` because I don't think it can influence the interpretation of the search term any more. There is now code in query/interpretation.rb that checks that a name matches the pattern `/\b_/` (eg `_left`, `_right`) before absolutizing.  So the search term won't get absolutized unless it's relative, which is already a weird case.

This means we can use, eg, `Company+*type+by_name` as an options card on autocomplete.

card/mod/core/set/all/haml.rb
```diff
@@ -2,22 +2,22 @@
...
 
-format do
   # Renders haml templates. The haml template can be passed as string or
   # block or a symbol that refers to a view template.
   # @param  template_or_locals [Hash, String, Symbol]
   #   If a symbol is given then a template is expected in the corresponding view
-  #   directory. Note that {view_template_path} needs to be overridden in
-  #   that case to get the right path to the template.
+  #   directory.
   # @return [String] rendered haml as HTML
```
ethn: I need to document the new pattern, which in short is this:
1. `haml` replaces `render_haml`
2. `haml_partial` replaces `render_haml_partial`
3. for `haml` the template path comes from the caller by default
4. for `haml_partial` (which should mostly be used within haml), the template path comes from the haml call that  called the partial
5. in either case, a `:template_path` argument may be sent in the locals to override the default path.

ethn: but I want to get your approval of that pattern first :)

card/mod/core/set/all/haml.rb
```diff
@@ -41,53 +41,38 @@
...
-      haml_to_html yield, template_or_locals, locals_or_binding
+
+  def haml *args, &block
+    if args.first.is_a? Symbol
+      process_haml_template *args
     else
-      haml_to_html template_or_locals, locals_or_binding, a_binding
+      process_haml *args, &block
     end
   end
```
ethn: much less brain bending :)

card/mod/machines/lib/javascript/decko_filter.js.coffee
```diff
@@ -23,25 +23,28 @@ filterCategorySelected = (addFilterDropdown, selectedCategory, label) ->
...
 
-
 showFilterInputField = (category, widget) ->
-  $searchInputField = $(widget.find("._filter-input-field-prototypes > ._filter-input-field.#{category} > .input-group")[0])
+  selector = "._filter-input-field-prototypes > ._filter-input-field.#{category} > .input-group"
+  $inputField = $(widget.find(selector)[0])
+
+  $(widget.find("._add-filter-dropdown")).before($inputField)
+  setFilterInputWidth $inputField
+  decko.initAutoCardPlete($inputField.find('input')) # only has effect if there is a data-options-card value
```
ethn: this is the only real change

card/mod/pointer/lib/javascript/script_pointer_config.js.coffee
```diff
@@ -24,12 +24,13 @@ decko.editorInitFunctionMap['.pointer-list-editor'] = ->
...
-    optionsCard ||= input.closest('.pointer-list-editor').data('options-card')
-    input.autocomplete {
-      source: decko.prepUrl decko.rootPath + '/' + optionsCard +
-          '.json?view=junction_name_complete'
-    }
+    decko.initAutoCardPlete input
+
+  initAutoCardPlete: (input) ->
+    optionsCard = input.data 'options-card'
+    return unless !!optionsCard
```
ethn: hmm.  better: `if !optionsCard`

card/mod/pointer/lib/javascript/script_pointer_config.js.coffee
```diff
@@ -24,12 +24,13 @@ decko.editorInitFunctionMap['.pointer-list-editor'] = ->
 
 $.extend decko,
   initPointerList: (input) ->
-    optionsCard = input.closest('ul').data('options-card')
-    optionsCard ||= input.closest('.pointer-list-editor').data('options-card')
-    input.autocomplete {
-      source: decko.prepUrl decko.rootPath + '/' + optionsCard +
-          '.json?view=junction_name_complete'
-    }
+    decko.initAutoCardPlete input
```
ethn: now we put the `options-card` class directly on the input, which makes it easier to reuse this pattern.

card/mod/pointer/set/abstract/02_pointer/options.rb
ethn: fixed.

card/mod/pointer/template/abstract/02_pointer/edit/list_item.haml
ethn: fixed

card/lib/card/set/format/haml_views.rb
xithan: It saves a line and a variable, is more reliable and looks better.

ethn: Oh, I agree. I was sure I already did this.  Must not have pushed.

ethn: wait, I did fix this. this is an outdated diff.

card/lib/card/set/format/haml_views.rb
xithan: The only reason I can think of why I added this is that somewhere it wasn't working without it. 
Maybe there are `render_haml` calls somewhere instead of a `render :view` call?

card/mod/pointer/lib/javascript/script_pointer_config.js.coffee
xithan: `unless optionsCard` doesn't work in coffeescript?

card/lib/card/set/format/haml_views.rb
xithan: If it works with wikirate I'm willing to take the risk, since I know now where I have to look if I encounter render issues with haml.

card/mod/pointer/lib/javascript/script_pointer_config.js.coffee
ethn: that is my memory, but I woudn't swear it.

PR #109, opened 2017-11-18
Subcards refactor

ethn:


card/lib/card/subcards/add.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      def new_by_attributes name, attributes={}
+        attributes ||= {}
+        absolute_name = absolutize_subcard_name name
+        subcard_args = extract_subcard_args! name, attributes
+        t_i_s = attributes.delete(:transact_in_stage)
+        card = initialize_by_attributes absolute_name, attributes
+        subcard = new_by_card card, transact_in_stage: t_i_s
+        card.subcards.add subcard_args
+        subcard
```
xithan: so how do you handle the add A+B+C to A case now

xithan: Is this tested somewhere that this really works?

PR #111, opened 2017-11-21
Nests

ethn:
edit_nests view was breaking.  led me into breaking up collection.rb


card/mod/core/set/all/csv.rb
```diff
@@ -0,0 +1,64 @@
+format :csv do
```
xithan: there is already an all/all_csv.rb file

card/mod/core/set/all/chunk.rb
```diff
@@ -0,0 +1,119 @@
+
+def chunks content, type
+  content ||= self.content
+  type ||= Card::Content::Chunk
+  Card::Content.new(content, self).find_chunks type
+end
+
+def each_chunk content, type
```
ethn: API change.  these were args before

card/mod/core/set/all/csv.rb
ethn: ok, I'll move these there.

card/mod/core/spec/set/all/collection_spec.rb
```diff
@@ -26,12 +26,12 @@
     end
   end
 
-  describe "#extended_list" do
-    it "returns item's content for pointer setting" do
-      c = Card.new(name: "foo", type: "Pointer", content: "[[Z]]")
-      expect(c.extended_list).to eq(["I'm here to be referenced to"])
-    end
-  end
+  # describe "#extended_list" do
```
ethn: removed API.  (not used in wikirate)

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -125,8 +125,8 @@ def editor_in_multi_card
     end
   end
 
-  def multi_card_edit_fields
-    nested_fields_for_edit.map do |name, options|
+  def multi_card_edit fields_only=false
```
ethn: #multi_card_edit handles all nests (fields or not) by default.  but the most prominent call is from edit_slot, which restricts to fields.

card/mod/standard/spec/set/all/rich_html/editing_spec.rb
```diff
@@ -17,6 +16,18 @@ def assert_active_toolbar_pill view, content, related_view=false
     end
   end
 
+  describe "edit_nests view" do
+    it "shows editors for both absolute and relative nests" do
```
ethn: this is what was broken and what started all this.

card/mod/core/set/all/extended.rb
```diff
@@ -0,0 +1,45 @@
+
+def extended_item_cards context=nil
```
ethn: significantly refactored.

PR #114, opened 2017-11-22
nested subfields

xithan:


card/lib/card/subcards.rb
```diff
@@ -236,11 +236,20 @@ def extract_subcard_args! name, args
       subcards
     end
 
+    # ensure a leading '+'
+    def normalize_subfield_key key
+      if key.is_a?(Symbol) && Card::Codename.exist?(key)
+        key = Card::Codename.name(key)
+      end
+      key = "+#{key}" unless key.to_s.start_with? "+"
```
ethn: `key.to_name.prepend_joint` doesn't require the conditional and avoids hard-coded plusses.

A shorter name would be better, since this is common.  `to_field` maybe?

card/lib/card/subcards.rb
```diff
@@ -236,11 +236,20 @@ def extract_subcard_args! name, args
...
+      key
+    end
+
     def new_by_attributes name, attributes={}
       attributes ||= {}
       absolute_name = absolutize_subcard_name name
       if absolute_name.field_of?(@context_card.name) &&
-         (absolute_name.parts.size - @context_card.name.parts.size) > 2
+         (absolute_name.parts.size - @context_card.name.parts.size) >= 2
```
ethn: I'm pretty sure this is never going to pass.  If `absolute_name.field_of?(@context_card.name)` is true, then the result of this subtraction will ALWAYS be exactly 1.

card/spec/lib/card/subcards_spec.rb
```diff
@@ -138,6 +138,26 @@
...
+    it "handles codenames" do
+      create "card with subs", subfields: { title: "title 1" }
+      expect_card("card with subs+*title").to exist.and have_db_content "title 1"
+    end
+
+    it "handles nested subfields" do
+      create "card with subs",
+             subfields: { "nested" => { subfields: { title: "title 2"} } }
+      expect_card("card with subs+nested+*title").to exist.and have_db_content "title 2"
+    end
```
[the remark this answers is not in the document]
xithan: yes. This bug is not related to the refactoring.

xithan:
An unpassable condition doesn't necessarily mean that the code is not needed. It's can also mean that only the condition is wrong.

PR #115, opened 2017-11-22
don't use git in gemspec

xithan:
[Finishes #153098287]

### DIFF
--- a/card/card.gemspec
+++ b/card/card.gemspec
@@ -28,24 +28,9 @@ Gem::Specification.new do |s|
   s.homepage      = "http://decko.org"
   s.licenses      = ["GPL-2.0", "GPL-3.0"]
 
-  s.files         = `git ls-files`.split $INPUT_RECORD_SEPARATOR
+  s.files         = Dir["VERSION", "README.rdoc", "LICENSE", "GPL",
+                        "{config,db,lib,mod,tmpsets}/**/*"]
 
-  # add submodule files (seed data)
-  morepaths = `git submodule --quiet foreach pwd`.split $OUTPUT_RECORD_SEPARATOR
-  morepaths.each do |submod_path|
-    gem_root = File.expand_path File.dirname(__FILE__)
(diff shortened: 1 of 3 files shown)


card/card.gemspec
```diff
@@ -28,24 +28,9 @@ Gem::Specification.new do |s|
...
-      s.files += morefiles.map do |filename|
-        "#{relative_submod_path}/#{filename}" #unless filename =~ /\s/
-      end.compact
-    end
-  end
-
+  s.files         = Dir["VERSION", "README.rdoc", "LICENSE", "GPL",
+                        "{config,db,lib,mod,tmpsets}/**/*"]
 
-  s.test_files    = s.files.grep(%r{^(test|spec|features)/})
```
ethn: any reason not to have `s.test_files`?

card/card.gemspec
```diff
@@ -28,24 +28,9 @@ Gem::Specification.new do |s|
...
-      morefiles = `git ls-files`.split /\n/ #$OUTPUT_RECORD_SEPARATOR
-      #puts "ors = #{$OUTPUT_RECORD_SEPARATOR}"
-      s.files += morefiles.map do |filename|
-        "#{relative_submod_path}/#{filename}" #unless filename =~ /\s/
-      end.compact
-    end
-  end
-
+  s.files         = Dir["VERSION", "README.rdoc", "LICENSE", "GPL",
+                        "{config,db,lib,mod,tmpsets}/**/*"]
```
ethn: This excludes dotfiles. I had a look to see if any needed to be in the gem.  I think `card/.yardopts` does, because sometimes yard docs are generated directly from the gem.  There's also `.yardoc` in the root directory, but I'm not sure it does anything.

xithan: I don't see a `.yardoc` in the root directory

card/card.gemspec
xithan: It seems to be deprecated and I don't see the value of adding test files to the gem

PR #121, opened 2017-11-24
css class to make arbitrary elements a slotter

xithan:
used for table rows in wikirate


card/mod/machines/lib/javascript/decko.js.coffee
```diff
@@ -194,6 +194,9 @@ $(window).ready ->
     return false if !$.rails.allowAction $(this)
     $.rails.handleRemote $(this)
 
+  $('body').on 'click', '._clickable.slotter', (event)->
+    $.rails.handleRemote $(this)
+
```
ethn: Should our policy be to update machine output in all PRs that change assets?

PR #122, opened 2017-11-24
 support non-fields in edit_fields

xithan:


card/mod/core/spec/set/all/chunk_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    it "interprets strings as field names" do
+      format = format_with_edit_fields ["B", "+C"]
+      expect(format.render_edit).to have_tag ".card-slot" do
+        with_tag ".card-editor.RIGHT-b", with: { card_name: "A+B" }
+        with_tag ".card-editor.RIGHT-c", with: { card_name: "A+C" }
+      end
+    end
+
+    it "doesn't make card objects to fields" do
```
ethn: If it's not only for fields, I think we should rename `#edit_fields`.  (I ran into the same issue in my refactoring.)

PR #124, opened 2017-11-27
tinymce config

xithan:
[Finished #153181469]

### DIFF
--- a/card/mod/tinymce_editor/db/migrate_cards/20171113233151_update_tinymce_config.rb
+++ b/card/mod/tinymce_editor/db/migrate_cards/20171113233151_update_tinymce_config.rb
@@ -4,6 +4,7 @@ class UpdateTinymceConfig < Card::Migration::Core
   TINYMCE_CONF = <<-JSON.strip_heredoc
     {
       "theme": "modern",
+      "menubar": "edit view insert format table",
       "plugins": "autoresize code lists hr link autolink table contextmenu textcolor colorpicker",
       "toolbar1": "formatselect | bold italic strikethrough forecolor backcolor | link | alignleft aligncenter alignright alignjustify  | numlist bullist outdent indent  | removeformat",
       "width":"100%",


card/mod/tinymce_editor/db/migrate_cards/20171113233151_update_tinymce_config.rb
```diff
@@ -4,6 +4,7 @@ class UpdateTinymceConfig < Card::Migration::Core
   TINYMCE_CONF = <<-JSON.strip_heredoc
     {
       "theme": "modern",
+      "menubar": "edit view insert format table",
```
ethn: can we either take table off this list or remove the stripping right away?  So long as they're both there the brokenness is pretty harsh.

xithan: We don't strip tables.

xithan: The only issue is if you set `Card.config.allow_inline_styles` to false then the format options for tables won't have any effect.

ethn: OK thanks, this is helping me remember the issues.

The whole `allow_inline_styles` thing was created because everybody hated all the problems we were having when cutting and pasting content.   Eg, you cut and paste content from one card to another, and you get weird nested card styling (at least in certain browsers) because the spreadsheet styles are translated into inline styles in the process.  Or you paste from nytimes and get their font.

docs.wikirate.eu allows inline styles, and content often gets messed up in ways that are hard to fix. It also shows that we don't have the styles we'd need to support tables well.  Eg http://docs.wikirate.eu/Project_milestones, in which the table juts out of the card, even if the browser is as wide as my laptop.

We don't want to go back to that frustration.  On the other hand, you could argue that the most frustrating thing about the UX is when you get it the way you like it in the editor and then more things get stripped after the save.  I think we probably need to make it so that the editor does most of the work of cleaning up pasted content - the "clear formatting" button should be a big help.  Ruby should enforce any rules that are critical to functionality/security, but in general I think we need to take a tack of helping users create clean things

I guess I'm ok with #1 for now, but I would anticipate problems.  I would be a weak vote for the status quo (no tables yet), but that's not listed as an option.

PR #126, opened 2017-11-29
fix session cards

xithan:
and remove remaining "btn-default" css classes


card/mod/carrierwave/set/all/file_utils.rb
```diff
@@ -5,17 +5,19 @@ def update_all_storage_locations
   end
 
   def delete_tmp_files_of_cached_uploads
-    disposable_attachments do |card, action|
+    cards_with_disposable_attachments do |card, action|
```
ethn: ...or actions even...

card/mod/carrierwave/set/abstract/attachment/storage_type.rb
```diff
@@ -97,6 +97,10 @@ def change_bucket_if_read_only?
   cloud? && bucket_config[:read_only] && attachment_is_changing?
 end
 
+def read_only?
+  web? || (cloud? && bucket_config[:read_only])
+end
```
ethn: makes sense.

card/mod/carrierwave/set/all/file_utils.rb
```diff
@@ -5,11 +5,20 @@ def update_all_storage_locations
...
+  end
+
+  def cards_with_disposable_attachments
     draft_actions_with_attachment.each do |action|
       # we don't want to delete uploads in progress
       next unless old_enough?(action.created_at) && (card = action.card)
-      card.delete_files_for_action action
-      action.delete
+      # we can't delete attachments we don't have write access to
+      next if card.read_only?
```
ethn: somewhat surprised that code climate complains about cognitive complexity here, but it would be dead easy to grok if it were:

```
draft_actions_with_attachment.each do |action|
  next unless disposable_action? action
  yield action
end
```

ethn: actually, you can just leave it at `old_enough?` if the above were:
```
disposable_attachment_actions do |action|
  next if action.card&.read_only?
  action.delete
end
```
This would change the logic slightly and delete actions without cards, but I'd consider that a good thing :)

card/mod/standard/set/type/session.rb
```diff
@@ -22,6 +22,10 @@ def content
   Env.session[key]
 end
 
+def content= val
+  Env.session[key] = val
+end
```
ethn: surprised we didn't have that!

card/lib/card/format/nest.rb
```diff
@@ -6,6 +6,13 @@ module Nest
       include Subformat
       include Mode
 
+      # @param cardish card mark
+      # @param options [Hash]
+      # @option options [Symbol] view
+      # @option options [Symbol, Array<Symbol>] :hide hide optional subviews
+      # @option options [Symbol, Array<Symbol>] :show show optional subviews
```
ethn: hide and show can also take single strings/symbols, no?

(Thanks for adding this!!)

PR #127, opened 2017-11-29
stop caching rules button

wagn:


card/mod/standard/set/self/rules_toolbar_button.rb
```diff
@@ -1,7 +1,7 @@
 include_set Abstract::ToolbarSplitButton
 
 format :html do
-  view :core do
+  view :core, cache: :never do
     subject.toolbar_split_button("rules", view: :edit_rules, icon: :list) do
```
wagn: This pattern – in which the view's subject is not the card itself – is never going to play well with view caching.

There can only be one cached version of a view for any given card.  In this case, there is only one card ever rendered: `:rules_toolbar_button`.   If this were done on the all set and rendered in a `rules_toolbar` view, caching should handle it just fine.

The most obvious other place where we're using this pattern is with the activity button, which has already been configured with `cache: :never`.

wagn: oops, I'm still wagn.  going back to ethn...

xithan: The extra card only for a button isn't a convincing solution but we don't have a better pattern established for this case. The problem is that we have to include an abstract set several times in the same set and override methods differently for different buttons. 
The best way to solve this is probably get rid of the abstract set and use ordinary ruby classes for every type of button.

xithan: Another guess: The buttons are cards so that you can remove them. But it doesn't work very well. You have to know that you have delete the codename first before you can remove the card and `return "" unless button_card = Card[button_codename]` will raise an exception without the codename.

ethn: I can imagine other reasons, too, like maybe we wanted permissions for different buttons. In any case, I think we agree that we haven't nailed the design yet.

xithan: It should be a virtual right set instead of a self set then we can have permissions and view caching.

ethn: I like it.

PR #128, opened 2017-11-30
refactor step files

xithan:
(cherry picked from commit 7d22cec)


xithan:
decko_steps.rb was the first one on codeclimate's list

PR #130, opened 2017-12-04
version 0.3.5 tick

ethn:


ethn:
these mandatory checks can be a pain for routine stuff.

PR #131, opened 2017-12-04
fix query option `append`

xithan:
#process_names was called with wrong number of arguments


card/spec/lib/card/query_spec.rb
```diff
@@ -24,6 +24,11 @@
       is_expected.to eq(%w(C+A D+A F+A))
     end
 
+    it "appends names to card results" do
+      res = Card::Query.run name: [:in, "C", "D", "F", "Z"], append: "A"
+      expect(res).to eq([Card["C+A"],Card["D+A"], Card["F+A"], nil])
+    end
```
xithan: Our query specs are insufficient. We only test name results.

card/lib/card/query/run.rb
```diff
@@ -72,7 +76,7 @@ def alter_results?
 
       def card_result record, _field
         if alter_results?
-          Card.fetch process_name(record["name"])
+          Card.fetch alter_result(record["name"])
         else
           fetch_or_instantiate record
```
xithan: Is there a reason that we don't cache when we alter the result?

ethn: #fetch does cache.  but we can't instantiate an altered result.  (all the information needed for normal instantiation is in the record, but the record isn't sufficient for an altered result

card/spec/lib/card/query_spec.rb
xithan: What should a search with an `append` statement return if the card doesn't exist? A new card or nothing? I think returning `nil`, what it currently does, will cause trouble somewhere someday.

ethn: There should be a `new: {}` there.  I broke it in this commit: https://github.com/xithan/decko/commit/6f1c02a8f7787ab0641385b5d398bd8a49ab86c4#diff-6b22a1f8bd11baaec00638be4a7d3814  (which was merged fairly recently)

ethn: (only needed in card_result.  `Card.fetch alter_result(record["name"]), new: {}`

PR #136, opened 2017-12-08
paging links fix

xithan:


card/lib/card/tasks/card/file_card_creator/abstract_file_card.rb
```diff
@@ -21,7 +21,7 @@ def initialize mod, name, type, codename: nil, force: false
         @type = type.to_sym
         @name = name
         @force = force
-        @codename = codename || name.to_name.key
+        @codename = codename || name.underscore.tr(" ","_")
```
ethn: why isn't the key better?

xithan: I mentioned the reason in the commit message: If the name is plural and you don't give an optional codename argument then you probably want the codename also be plural.

xithan: "I don't want the '*' of the name in the codename I add an extra codename argument" seems more intuitive to me than "The codename should also be plural like the name. I have to repeat it in the codename argument".

ethn: I'm sorry, I misunderstood the code.  (and should have read the commit message).  all sounds good.

PR #143, opened 2017-12-22
Calcs

ethn:
changes for new calculated metric handling.


card/mod/core/set/all/name_validations.rb
```diff
@@ -5,8 +5,8 @@
 end
 
 event :validate_uniqueness_of_codename do
-  return unless codename.present? && errors.empty? &&
-                Card.find_by_codename(codename).present?
+  return (self.codename = nil) unless codename.present?
+  return unless errors.empty? && Card.find_by_codename(codename).present?
```
xithan: I find the negated versions easier to read, e.g.`if errors.present? || Card.find_by_codename(codename).blank?`

card/lib/card/query/value.rb
```diff
@@ -14,18 +15,27 @@ def initialize rawvalue, query
...
         when String, Integer then ["=", rawvalue]
-        when Array           then [rawvalue[0], rawvalue[1..-1]]
+        when Array           then parse_array_value rawvalue
         else raise("Invalid Condition Clause #{rawvalue}.inspect}")
         end
       end
 
+      def parse_array_value array
+        operator = operator?(array.first) ? array.shift : :in
+        [operator, array]
```
ethn: this is where the change is.

xithan: Doesn't matter anymore but what's the difference between `[rawvalue[0], rawvalue[1..-1]]` and `rawvalue`?

ethn: it's the difference between `[op, v1, v2]` and `[op, [v1, v2]]`.  definitely easier to see that now! :)

PR #144, opened 2018-01-08
deck structure migrations

xithan:


card/config/initializers/02_patches/active_record.rb
```diff
@@ -104,5 +104,18 @@ def match string
         end
       end
     end
+
+    module Migration
+      module ClassMethods
+        def check_pending!(connection = ::ActiveRecord::Base.connection)
+          [:structure, :core_cards, :deck, :deck_cards].each do |migration_type|
```
ethn: just to have written this somewhere, I think the better names would be `core_structure`, `core_cards`, `deck_structure`, and `deck_cards`. 

But to make this work, we might have to migrate the name of the schema migration tables :)

decko/lib/decko/config/environments/cucumber.rb
```diff
@@ -10,6 +10,7 @@
   # For more information see https://rspec.lighthouseapp.com/projects/16211/tickets/165
   config.cache_classes = true
 
+  config.persistent_cache = false
```
ethn: upon further review, I think this is wrong.  it looks like cucumber was in fact using the persistent cache.  (https://github.com/decko-commons/decko/commit/75d4f685c05193fc6995a83801e1f5703a07d02e#diff-84503ec45e737516bfc7987cacc5839f)

card/mod/search/set/abstract/wql_search.rb
```diff
@@ -8,7 +8,12 @@ def search args={}
   query.run
 end
 
+def cache_query?
+  true
+end
```
ethn: I assume there is a new wikirate use case where we don't want to cache the query?

xithan: It's an old wikirate use case that you broke with your search refactoring a while ago (when you introduced `wql_from_content` and started to cache the return value of `wql_hash`) but it didn't break any tests until I had to regenerate the seed data while I was struggling with a mess of failing tests.

PR #145, opened 2018-01-09
make paging for pointers work again

xithan:


card/mod/pointer/set/abstract/02_pointer.rb
```diff
@@ -54,7 +54,7 @@ def pointer_items args={}
 end
 
 format :html do
-  view :core do
+  view :core, cache: :never do
```
ethn: kind of a bummer to lose caching here, but I'm not immediately thinking of a better solution.

we could only have paging in a separate (non-core) view, but that means no paging in content, titled, open, etc.

we could have improve the caching API, eg so that you can prevent caching if paging is on, but that's a headache and adds complexity.

so...I approve :)

ethn:
Did Donald Trump name this PR?

PR #146, opened 2018-01-12
card load hooks

xithan:
I need that to add new relic traces to events


card/lib/card.rb
```diff
@@ -1,7 +1,7 @@
 # -*- encoding : utf-8 -*-
-require "carrierwave"
 
 Object.const_remove_if_defined :Card
+ActiveSupport.run_load_hooks(:before_card, self)
```
ethn: I like it.

PR #147, opened 2018-01-12
Ethn wikirate

ethn:
moves out wikirate-specific filter formgroups, wraps filter_items in a slot (which fixes several problems)


card/lib/card/set/event.rb
```diff
@@ -12,69 +12,93 @@ module Event
...
-      def define_event_method event, call_method
+      # EVENT DEFINITION
+
+      def define_event event, opts, &final
+        simple_method_name = "#{event}_without_callbacks"
+        define_simple_method simple_method_name, &final
+        define_event_method event, simple_method_name, opts
+      end
+
+      def define_simple_method method_name, &method
```
xithan: you have to adjust the new_relic mod in card-mods when you rename this

ethn: thanks.  good catch.

PR #148, opened 2018-01-14
use firefox instead of chrome for cukes

xithan:
The current version of selenium and geckodriver seems to work more reliable with firefox.


xithan:
semaphore is not happy. Its firefox is too old.

PR #151, opened 2018-01-20
version tick (and minor api tweak)

ethn:


card/lib/card/set/event.rb
```diff
@@ -59,11 +59,11 @@ def callback_name stage, after_subcards=false
 
       def define_event event, opts, &final
         simple_method_name = "#{event}_without_callbacks"
-        define_simple_method simple_method_name, &final
+        define_simple_method event, simple_method_name, &final
         define_event_method event, simple_method_name, opts
       end
 
-      def define_simple_method method_name, &method
+      def define_simple_method _event, method_name, &method
```
ethn: needed for new_relic mod

xithan:
maybe include #152 in this version tick?

ethn:
I already pushed the gem and the release for 0.3.7.  But I'm happy to do 0.3.8

xithan:
You pushed a gem that didn't pass CI? 😟

ethn:
At the time that I pushed it, the version itself was the only thing that
hadn't passed CI.  At that point CI was breaking everywhere, including on
master, because of the Firefox thing.

I've added a few more tweaks since, but those aren't in the release.

ethn:
At the time that I pushed it, the version itself was the only thing that
hadn't passed CI.  At that point CI was breaking everywhere, including on
master, because of the Firefox thing.

I've added a few more tweaks since, but those aren't in the release

PR #154, opened 2018-01-25
large list logging / rss opt-in

ethn:
another attempt to get information about server problems.


card/mod/basic_formats/set/self/head.rb
```diff
@@ -34,7 +34,7 @@ def head_buttons
     if root.card
       bits << universal_edit_button
       # RSS # move to mods!
-      bits << rss_link if root.card.type_id == SearchTypeID
+      bits << rss_link
```
ethn: rss_link_tag moved to search type (imo, these type_id `if` tests are often a red flag)

card/mod/standard/spec/set/all/rich_html/form_elements_spec.rb
```diff
@@ -18,5 +18,19 @@ def hidden_tags opts
...
+
+      example "first array value" do
+        expect(tags)
+          .to have_tag(:input, with: { type: "hidden", name: "a[b][]", value: "1" })
+      end
+
+      example "second array value" do
+        expect(tags)
+          .to have_tag(:input, with: { type: "hidden", name: "a[b][]", value: "2" })
+      end
```
ethn: these are kinda lame when separated, but codeclimate made me do it.

PR #157, opened 2018-02-08
overlay

xithan:


card/mod/standard/set/all/rich_html/overlay.rb
```diff
@@ -0,0 +1,38 @@
+format :html do
+  view :overlay do
+    overlay [_render_open_content, render_comment_box]
+  end
+
+  def overlay content=nil
+    class_up "d0-card-frame", "card"
+    class_up "card-slot", "_overlay bg-white", true
+    @content_body = true
+    frame do
```
ethn: why not just call `overlay_frame` directly?

xithan: I removed `overlay_frame`

Issue #149, opened 2018-01-18
More issues with nginx


Nemquae:
Hey, I'm finally getting around to setting up a server using the latest decko, and I'm running into some issues with nginx here.  I've set the file/folder permissions to 777 (just to get this up an running), but it still can't seem to read/write several important files on load.

Here's a screenshot:
![decko testing](https://user-images.githubusercontent.com/237217/35096413-10eed680-fc87-11e7-9d7e-f35e6ebc1107.png)

And here are the log files:
[production.log](https://github.com/decko-commons/decko/files/1642695/production.log)
[development.log](https://github.com/decko-commons/decko/files/1642696/development.log)

I've never set up a server like this before and we need this soon, so any help is appreciated.

ethn:
Are you trying to run this with a relative root (gamesdb)?

Because if so you will need to configure that.

Here's the relevant Rails documentation for how to do that: http://guides.rubyonrails.org/configuring.html#deploy-to-a-subdirectory-relative-url-root

Short version, you have some options:
config.relative_url_root = "/gamesdb" (eg in config/application.rb)
RAILS_RELATIVE_URL_ROOT = "/gamesdb" (as an environmental variable)
in your passenger configuration (.htaccess): RackBaseURI = "/gamesdb"

more recent passenger configuration may be using a different option, but it's something like that...

Nemquae:
Yes, and that fixed it (mostly).

The last part we're stuck on are the icons (and I presume other assets) that are still looking for the root.  Everything else about the site seems to be working.

Here are the errors:

> MaterialIcons-Regular.woff2 Failed to load resource: the server responded with a status of 404 (Not Found)
> ui-bg_flat_75_ffffff_40x100.png Failed to load resource: the server responded with a status of 404 (Not Found)
> MaterialIcons-Regular.woff Failed to load resource: the server responded with a status of 404 (Not Found)
> MaterialIcons-Regular.ttf Failed to load resource: the server responded with a status of 404 (Not Found)
> DevTools failed to parse SourceMap: http://learning.sd.polyu.edu.hk/gamesdb/files/:all_script_machine_output/bootstrap.min.js.map
﻿
We tried clearing the cache and restarting nginx/passenger a few times, no luck.  Also, after we added memcached it's still not fast enough.  I know you mentioned xsendfile and expires, so we're looking into those, but were there any other nginx specific optimizations that you could recommend?

ethn:
[...]

PR #165, opened 2018-02-16
new decko.path handling in coffee

ethn:
`decko.path` now uses `decko.rootPath` (so you don't generally need to use rootPath directly)
`decko.prepUrl` is now `decko.slotPath`, which is a better name, because it returns a path with slot params. 

Since `slotPath` now uses `path`, it too handles relative roots.


card/mod/machines/lib/javascript/decko_slot.js.coffee
```diff
@@ -55,9 +56,8 @@ jQuery.fn.extend {
     $slot = $(this)
     $slot = $slot.slot() unless $slot.isSlot
     unless url?
-      url = decko.rootPath + '/~' + $slot.data('cardId') + "?view=" +
-        $slot.data("slot")["view"]
-      url = decko.prepUrl url, $slot
+      path = '~' + $slot.data('cardId') + "?view=" + $slot.data("slot")["view"]
+      url = decko.slotPath path, $slot
```
ethn: this is a pretty good example of how much cleaner the new API is.

card/mod/tinymce_editor/lib/javascript/script_tinymce_config.js.coffee
```diff
@@ -33,7 +33,7 @@ $.extend decko,
       entity_encoding: 'raw'
     }
     $.extend conf, user_conf, hard_conf
-    tinyMCE.baseURL = '/assets/tinymce'
+    tinyMCE.baseURL = decko.path('assets/tinymce')
```
ethn: this fixes the tinyMCE bug in https://github.com/decko-commons/decko/issues/149

decko/README.rdoc
```diff
@@ -81,7 +81,23 @@ Run the following:
 
 ====4. Restart your server.
 
-===Upgrading pre-gem Wagn sites
+===Upgrade to Decko from Wagn
```
ethn: just realized we didn't have this anywhere.

ethn: technically we could automate some of this, but running `decko update` would break unless most of it was already done...

card/mod/machines/lib/javascript/decko_filter.js.coffee
```diff
@@ -47,14 +47,14 @@ $(window).ready ->
     $(this).closest('li').remove()
 
 newFilteredListContent = (el) ->
-  decko.pointerContent prefilteredNames(el).concat(selectedNames el)
+  $.map(prefilteredIds(el).concat(selectedIds el), (id) -> "~" + id).join "\n"
```
ethn: this is the change that gets the filtered list working

PR #173, opened 2018-02-27
csv bugfix

xithan:


card/mod/basic_formats/set/all/all_csv.rb
```diff
@@ -9,11 +9,11 @@ def default_item_view
     depth.zero? ? :csv_row : :name
   end
 
-  view :core do
+  view :core do |args|
     if (item_view_options[:view] == :name_with_fields) && focal?
       title_row("item name") + name_with_field_rows
     else
-      super()
+      super(args)
```
ethn: I haven't looked at cukes, but I'm guessing this is why they're failing. (especially now that I notice only 3 lines changed).

The plan (or least the plan that has my vote) is to get rid of these args, and this is one part of the reason; the inconsistency screws up super.

As a short-term fix, you might be able do something like this:

`super.arity.zero? ? super() : super(args)`

..but I'm not sure super responds to `arity`.

Better would be to get rid of "args" in all core views, but I know that could be a big chunk of work.

PR #175, opened 2018-02-28
relative root fixes (needed for UBOS)

ethn:


card/VERSION
```diff
@@ -1 +1 @@
-0.3.11
```
ethn: seems like time for a minor release.  no huge reason for it to be this one...

decko/lib/decko/generators/decko/templates/config/routes.erb
```diff
@@ -1,12 +1,5 @@
...
-  rel_root = Cardio.config.relative_url_root
-<% if @include_jasmine_engine %>
-  if !Rails.env.production? && Object.const_defined?( :JasmineRails )
-    engine = Object.const_get(:JasmineRails).const_get :Engine
-    mount engine => "#{rel_root}/specs"
-  end
-<% end %>
-  mount Decko::Engine => "#{rel_root}/"
+  mount Decko::Engine => "/"
 end
```
xithan: Do you think we will find this in the git logs when some day we try to get JS tests to work again?

ethn: I can restore that. My priority was cleaning up the default routes.

PR #179, opened 2018-03-04
Update readme

chuenlok:
Add a few FAQ questions and a link to the FAQ in the `readme.rdoc`. I hope this can save people's time :)


FAQ.md
```diff
@@ -0,0 +1,25 @@
+# FAQ
```
ethn: Thanks so much for starting this!

I think we're going to want to divide our documentation up into four main use cases:

A. Deckering (wagneering)
B. Installation / server configuration
C. Mod Development
D. Core Development

For each of these, we need to decide whether the primary documentation is in the code or on wagn/decko.org (we'll be switching to decko.org very soon, I hope).

(A) Easy, these folks may never look at GitHub, so all documentation for them should be on decko.org
(B) This is the trickiest one.  We currently have a some duplication (eg the basic installation directions appear on both), but I'm think decko.org needs to be the go-to place. We have a lot more community-generated content there (eg OS-specific stuff), there will be some installations (like UBOS) where people don't interact directly with code/github, and 
(C) probably code, though perhaps they will use it most on somewhere like this: http://www.rubydoc.info/gems/card/Card. Basically, it will be the equivalent of api.rubyonrails.org.
(D) definitely code

FAQ.md
```diff
@@ -0,0 +1,25 @@
...
+   ```
+
+   This command will then ask for the local decko path. The created deck will be using your local `decko` folder as the gem path in the `GemFile`.
+
+1. **While I started a decko server with the local decko path, it showed the error of `NoMethodError (undefined method 'notable_exception_raised' for #<Card:0x007f8d44fed250>):`.**
+
+   After cloning the decko code from github, there are serval submodules that need to be initialized. You may run the command `git submodule update --init --recursive` in your local decko root directory. Then redo the seeding `bundle exec decko seed`.
+
+1. **I have a problem to upload image to my site. The image uploaded is shown as zero bytes with a broken image in the preview but uploading a non-image is fine.**
+
```
ethn: This is in group (B), so we need to get it on wagn.org.  We also need to make sure that github points people to decko.org for troubleshooting.

FAQ.md
```diff
@@ -0,0 +1,25 @@
...
+   ```
+   $ decko new <Deck Name> --core-dev
+   ```
+
+   This command will then ask for the local decko path. The created deck will be using your local `decko` folder as the gem path in the `GemFile`.
+
+1. **While I started a decko server with the local decko path, it showed the error of `NoMethodError (undefined method 'notable_exception_raised' for #<Card:0x007f8d44fed250>):`.**
+
+   After cloning the decko code from github, there are serval submodules that need to be initialized. You may run the command `git submodule update --init --recursive` in your local decko root directory. Then redo the seeding `bundle exec decko seed`.
+
```
ethn: I tried to replicate this and couldn't.  So did you have NO code or OLD code in that subrepository?

For now, I'm thinking this should go in a "Troubleshooting" section at the bottom of Contributing.rdoc.  (and we should make sure the README links to the Contributing doc!

decko/README.rdoc
```diff
@@ -123,4 +123,7 @@ If you're new to Decko, you'll find lots more info at http://wagn.org
 If you're looking to go deep into the code, start with http://www.rubydoc.info/gems/card
 
 
+==FAQ
+You may find some useful information here[link:FAQ.md] when you encountered problems.
```
ethn: I think we need four things in the More Resources section that align with the four groups:
A: wagn.org for general resources
B: wagn.org/installation for installation problems (need to be able to find the imagemagick stuff from there)
C. rubydoc.info/gems/card for mod developers
D. Contributing.md for core developers.

chuenlok:
Thanks Ethan! I will update it later in this week :)

I don't have access to the wagn.org or decko.org so I think I will write in github first and you can convert it and put in wagn/decko.org later

ethn:
That sounds great. 

I do think you have a wagn.org account associated with your gmail address (http://wagn.org/Henry_Tai). But do whatever is most comfortable.

I'm with Vignesh and Philipp in Berlin at the moment. We're missing you!

FAQ.md
chuenlok: I do not have any code in the subrepository at the beginning and the error message is not leading to the missing submodule code to a beginner.

chuenlok:
I missed you guy too! Hope your next meet will be in the UK and we can meet!

I made some changes but not much. I hope it is still useful. 

About the error I had with missing submodules, I do not have any code in the submodule directory so the error is shown. I think most of the beginner will face the problem as the error is not intuitive to the problem.

FAQ.md
ethn: Yes, I definitely want to make this easy to find.

My main concern is that we separate developer beginners from "normal" beginners.  Normal beginners are installing from gems, so they won't have the submodule problem. Only developers will install the code.

Right now we barely have any documentation for new mod or core developers.  That's a real need!

ethn:
Thanks, Henry.

I'm going to work on docs and get this merged in soon!

PR #185, opened 2018-03-15
no schema dumping

ethn:


decko/lib/decko/config/environments/production.rb
```diff
@@ -65,5 +65,6 @@
   # Send deprecation notices to registered listeners
   config.active_support.deprecation = :notify
 
-  ENV["RAILS_ASSET_ID"] = ""
```
ethn: I couldn't find any references to this in the rails codebase any more.

decko/lib/decko/config/environments/production.rb
```diff
@@ -65,5 +65,6 @@
...
+  # do not regenerate schema.rb
+  config.active_record.dump_schema_after_migration = false
```
xithan: Don't we want this disabled for the other environments, too?

PR #186, opened 2018-03-16
name escape

xithan:
escape <, >, and / in card names


cardname/lib/cardname.rb
```diff
@@ -59,7 +58,9 @@ def stringify obj
     end
 
     def banned_re
-      %r{#{ (['['] + banned_array << joint) * '\\' + ']' }}
+      Regexp.escape joint
```
ethn: seems like if we stop supporting the banned_array, we should take it out completely.  Now it looks like we support it but we don't.

PR #187, opened 2018-03-19
query params in params[:query]

ethn:


xithan:
don't forget to change wikirate!

PR #188, opened 2018-03-19
optional events

xithan:
API to skip events


card/mod/core/set/all/event_conditions.rb
```diff
@@ -56,3 +61,8 @@ def wrong_action action
   return false if on_condition_applies? action
```
ethn: need another argument here.

PR #193, opened 2018-03-30
getting rid of view args

ethn:
(now ready for merge)
This work all started because of (yet another) caching problem related to view args.  I decided to take the jump and try to get us to where we can eliminate the args altogether. Getting rid of view args will allow us to simplify the view api considerably:

1. The view API will _only_ support known options. So (once this is finished) if you try to call render with an unknown option, it will fail immediately (as opposed to emerging as a bug in peculiar caching conditions).
2. The above simplifies caching considerations, as it was very hard to explain (and remember) how the foreign options broke things. 
3. Generally speaking, this forces cleaner, more deckster- and caching- friendly views.

There are still a (very) few views still using args. To finish this work, I'm also thinking:
1. to add explicit param support to views.  Ie, a view has to specify the environmental params it considers. This will actually make more views cacheable (because the param value will be in the stub)
2. relatedly, you should be able to set these params for the view context as a render arg.  Eg.  `render :viewname, params: { metric_name: 'my metric' }`,  This will set the param for the environment of the view (but not alter the whole environment!).  We need this pattern for related views, imo.
3. `default_viewname_args args` should become `before_viewname`.  (no args involved. it's just for manipulating the voo)
4. `live_options` should be a private method within Card::View.  mod developers should not directly manipulate that hash.  Too hard to explain and too easy to create problems.
5. `voo.closest_live_option :option` should be replaced with `voo.inherit :option`.  Same concept, but easier to explain.  And it will be limited to the pre-defined options list.

### DIFF
--- a/card/mod/solid_cache/set/abstract/solid_cache.rb
+++ b/card/mod/solid_cache/set/abstract/solid_cache.rb
@@ -78,8 +78,8 @@ module ClassMethods
 end
 
 def with_solid_cache
-  card.update_solid_cache if card.solid_cache_card.new?
-  yield card.solid_cache_card
+  update_solid_cache if solid_cache_card.new?
+  yield solid_cache_card
 end
 
 def expire_solid_cache _changed_card=nil


card/mod/core/set/all/trash.rb
```diff
@@ -58,24 +58,29 @@ def delete!
...
 end
 
+def pull_from_trash!
+  return unless (trashed_card = Card.find_by_key_and_trash key, true)
+  # fwiw, now we _could_ get card using fetch look_in_trash: true (not tried).
+
+  self.id = trashed_card.id
+  # following is needed so that #id_in_database returns existing card id
+  # (and record is updated correctly)
+  db_attributes["id"] = trashed_card.db_attributes["id"]
```
ethn: this is the 5.1.6 fix

ethn:
This work pushed me to do some refactoring of:

1. history/act/action views.  I fixed several bugs along the way (eg bad numbering), got rid of unused methods (unused by decko or wikirate!), and generally made stuff cleaner and more readable. Should be a good start towards the new history handling we want.
2. exports.  There is much less duplication now, as you only have to specify "items_for_export" in different sets, and the export card does all the cleaning.
3. solid cache. old model used an arg-based conditional override of core view.  now it requires a "cacheable core" view. 
4. following/notifications.  no major reconceptualization, but there were lots of long/complex views.

card/mod/bootstrap/set/all/bootstrap/accordion.rb
```diff
@@ -1,5 +1,6 @@
 format :html do
-  def accordion_group list, collapse_id=nil, args={}
+  def accordion_group collapse_id=nil, args={}
+    list = yield
```
xithan: 👎

card/mod/core/set/all/export.rb
```diff
@@ -1,42 +1,51 @@
...
-    )
+  def default_export_items_args _args
+    @exported_keys ||= ::Set.new
   end
 
-  def default_export_items_args args
-    args[:processed_keys] ||= ::Set.new
+  view :export_items, cache: :never do
+    valid_items_for_export.map do |item|
+      subformat(item).render_export
```
xithan: `subformat`? not `nest`? In my experience `subformat` calls cause `voo` troubles

card/mod/search/set/type/search_type.rb
```diff
@@ -20,20 +20,8 @@ def search_result_view
 end
 
 format :json do
-  view :export do |args|
-    # avoid running the search from options and structure that
-    # case a huge result or error
```
xithan: no longer a problem?

card/mod/solid_cache/set/abstract/solid_cache.rb
```diff
@@ -1,7 +1,7 @@
 # -*- encoding : utf-8 -*-
 
-# A card that includes Abstract::SolidCache caches its core view
-# in a '+*solid cache' card.
+# A card that includes Abstract::SolidCache has as its core view a cached version of
+# its "cacheable_core" view, which is storeed in a '+*solid cache' card.
```
xithan: How can I cache the core view of a card with this approach?

card/mod/standard/set/type/session.rb
```diff
@@ -27,7 +27,7 @@ def content= val
 end
 
 format :html do
-  def default_core_args args
-    args[:items] = { view: :name }
+  def default_core_args _args
+    voo.items[:view] = :name
```
xithan: Are you sure `voo.items` is always initialized?

card/spec/support/helper/render_helper.rb
```diff
@@ -14,10 +14,10 @@ def render_content content, format_args={}
         render_content_with_args content, format_args
       end
 
-      def render_content_with_args content, format_args={}, view_args={}
+      def render_content_with_args content, format_args={}
         @card ||= Card.new name: "Tempo Rary 2"
         @card.content = content
-        @card.format(format_args)._render :core, view_args
+        @card.format(format_args)._render :core
```
xithan: I don't see a reason to keep this method.

card/mod/bootstrap/set/all/bootstrap/accordion.rb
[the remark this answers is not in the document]
ethn: see `act_accordion` for why I did this.  But I'm ok to go back.

card/mod/search/set/type/search_type.rb
ethn: hmm, I must have gotten confused going back and forth between this and the wikirate version.  See https://github.com/wikirate/wikirate/pull/791/files#diff-89b57322dad210a7070c1ab0ddd00c9f

I'll move that code here (and get rid of the wikirate file altogether)

card/mod/solid_cache/set/abstract/solid_cache.rb
ethn: you could do `view :cacheable_core, :core`.

I'm open to other solutions, so long as we get rid of the view args.

btw, I'm thinking we move this out of decko core and into a mod.  Seems like a clean one.

card/mod/standard/set/type/session.rb
ethn: it's a method defined in voo_api.rb

card/spec/support/helper/render_helper.rb
ethn: good point.  removing.

card/mod/core/set/all/export.rb
ethn: I agree. I think I was just subtracting and subtracting from the old method (which used subformat).  will fix.

card/mod/history/set/all/act_view.rb
```diff
@@ -1,39 +1,98 @@
-ACTS_PER_PAGE = Card.config.acts_per_page
+ACTS_PER_PAGE = 3 # Card.config.acts_per_page
```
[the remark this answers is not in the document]
ethn: because I was screwing around in the browser and forgot to put it back.

card/mod/bootstrap/set/all/bootstrap/accordion.rb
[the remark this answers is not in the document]
ethn: I'm not sure I understand option B, and I don't think it's a prominent enough method to warrant a complex API (option A).  So I'll just revert.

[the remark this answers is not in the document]
ethn: oh, I understand option B now :)

card/mod/solid_cache/set/abstract/solid_cache.rb
ethn: if you do `view :a, :b`, I don't think view a calls `render :b`; it defines view a using the block that defines view b.  You may have to specify the module where it's defined...

ethn: I came up with a more decksterlicious solution.

ethn:
Are you in agreement with the API changes proposed in the first comment?

PR #194, opened 2018-04-03
tweak name api

xithan:


card/lib/cardio.rb
```diff
@@ -50,7 +50,7 @@ def default_configs
         paging_limit:           20,
 
         non_createable_types:   [%w(signup setting set)],
-        view_cache:             true,
+        view_cache:             false,
```
ethn: If we change this, we have to make sure we change the config on staging/production.

I think I still favor defaulting to having it on, but I understand why you'd want it off.

card/lib/card/mark.rb
```diff
@@ -0,0 +1,64 @@
...
+      case mark
+      when ID_MARK_RE       then Regexp.last_match[:id].to_i
+      when CODENAME_MARK_RE then Card::Codename.id! Regexp.last_match[:codename]
+      end
+    end
+
+    def bad_mark mark
+      case mark
+      when ID_MARK_RE
+        raise Card::Error, "id doesn't exist: #{Regexp.last_match[:id]}"
```
ethn: Perhaps these should be NotFound?

card/lib/card/name.rb
```diff
@@ -34,19 +34,21 @@ def new str, validated_parts=nil
         str = str.to_s
         if !validated_parts && str.include?(joint)
           compose Cardname.split_parts(str)
-        elsif special_prefix?(str)
-          Card.fetch_name str # handles ~ and :
+        elsif (id = Card.id_from_string(str))  # handles ~ and :
+          Card.fetch_name(id) { Card.bad_mark(str) }
```
ethn: it makes me a little uncomfortable that we raise such a loud error from bad data.  But I'm not sure I have a better idea...

PR #195, opened 2018-04-05
fix handling of extensions in paths

ethn:


card/mod/core/set/all/references.rb
```diff
@@ -38,7 +38,7 @@ def name_referers
 # cards that refer to self or any descendant
 def family_referers
   @family_referers ||= ([self] + descendants).map(&:referers).flatten.uniq
-  # FIXME: could be much more efficient!
+  # TODO: make this more efficient!
```
ethn: I want to go through and fix "FIXME's" that are actually broken

card/mod/standard/set/all/links.rb
```diff
@@ -1,88 +1,66 @@
...
-    when /^https?\:/          then "external-link"
-    when %r{^/}               then "internal-link"
-    when /^mailto\:/          then "email-link"
-    when RESOURCE_TYPE_REGEXP then Regexp.last_match(1) + "-link"
-    end
+  # link to a different view of the current card
+  # @param view [Symbol,String]
+  # @param text [String]
+  # @param opts [Hash]
+  def link_to_view view, text=nil, opts={}
```
ethn: the main difference is that these methods now have their html/other format handling much more cleanly delineated.

PR #197, opened 2018-04-09
fixed issue with templates in structure views

ethn:
I know there aren't tests, but I think all of this is going to get refactored away soon, no?


ethn:
I decided to add a test after all.

PR #198, opened 2018-04-10
pointer bug fix

ethn:


card/mod/pointer/set/abstract/02_pointer/edit.rb
```diff
@@ -27,7 +27,7 @@ def add_item_button
   end
 
   def edit_list_item item
-    haml :list_item, pointer_item: item, options_card: options_card_name
+    haml :list_item, item: item, options_card: options_card_name
```
ethn: this was a screwup of mine from the view args cleanup. breaks all pointer editing.

card/mod/standard/set/all/rich_html/menu.rb
```diff
@@ -82,7 +82,7 @@ def menu_item_list link_opts={}
   end
 
   def menu_item_edit opts
-    menu_item "edit", "edit", opts.merge(view: :edit, path: menu_path_opts)
+    menu_item "edit", "edit", opts.merge(view: :edit)
```
ethn: this overrides the javascript slot handling, which was preventing edit view from holding on to slot item specifications.  That's why cancel was coming back in the wrong view on metric properties (it was actually the item that was in the wrong view(

xithan: and removing it doesn't cause any issues? I don't believe that.

ethn: I think `menu_path_opts` is a misleading method name; it's not meant to be used on menu items, only on the menu_link. (Ie, the name should probably be `menu_link_path_opts`). Everything in `menu_path_opts` is already handled in javascript, but presumably the `menu_link` needs special handling because it doesn't have normal slotter behavior.  (But it looks to me like it has more info than the menu_link needs, presumably because it was hacked to make the edit button work better).

That method (or, technically, a method that overrides it) is still called by `menu_item_edit` in `abstract/metric_answer/research.rb` in wikirate. I didn't want to rename it until I studied that issue.  My assumption is that we want the edit link (not the menu link) to have the RESEARCH_PARAMS_KEY handling.

ethn: Here's where you added `menu_path_opts`: https://github.com/decko-commons/decko/commit/d3c686f06dcaaa2177cf7504c4d42c2a7b682297

 It must mean something was wrong with the slot handling at the time.  But this is not the right solution to this problem. Every single link in the menu needs to have all that information and more.

xithan: maybe this is related 
higher_education_statistics_agency__hesa__research_income_-_wikirate

ethn: not related.  issue is this view in toolbar.rb:
```
    view "#{viewname}_link" do
      voo.title ||= viewtitle
      link_to_view viewname, voo.title, class: "dropdown-item"
    end
```
So these menu items break anywhere that the title is set.
I assume there are contexts where we actually need to set that link text, so it's probably not as simple as removing the voo.title stuff altogether...

ethn: ..but I can't find anywhere this is used, and I think, in principle, if we're using "title" to mean something other than "the title of the card in this context", we're asking for trouble.

PR #202, opened 2018-04-16
Related

ethn:
This turned into more than just the "related" view.  It stops supporting args in view definitions and disallows foreign view options.

To do that, I had to refactor `email_config` considerably.

I added a few notes in here, and there are also relevant comments in the wikirate PR.


card/config/initializers/01_core_extensions/hash.rb
```diff
@@ -52,6 +52,28 @@ def string_merge! other_hash, separator=" "
...
+          self[key].bury(*array)
+        end
+        self
+      end
+
+      private
+
+      def validate_bury_array! array
+        return unless array.find { |a| a.is_a? Hash }
+        raise StandardError, "no hashes in bury arguments"
```
xithan: `ArgumentError` seems to be a better fit

card/config/initializers/01_core_extensions/hash.rb
```diff
@@ -52,6 +52,28 @@ def string_merge! other_hash, separator=" "
...
+      # eg. hash.bury(:a, :b, 7) will make sure hash[:a][:b] is eq
+      # @return [Hash] with arbitrary depth
+      def bury *array
+        validate_bury_array! array
+        key = array.shift
+        if array.size == 1
+          self[key] = array.first
+        else
+          self[key] ||= {}
+          self[key].bury(*array)
```
xithan: you validate again with this call

card/config/initializers/01_core_extensions/hash.rb
```diff
@@ -52,6 +52,28 @@ def string_merge! other_hash, separator=" "
...
+      # opposite of #dig
+      # eg. hash.bury(:a, :b, 7) will make sure hash[:a][:b] is eq
+      # @return [Hash] with arbitrary depth
+      def bury *array
+        validate_bury_array! array
+        key = array.shift
+        if array.size == 1
+          self[key] = array.first
+        else
+          self[key] ||= {}
```
xithan: what if `self[key]` is true?

card/config/initializers/01_core_extensions/hash.rb
ethn: good point

card/config/initializers/01_core_extensions/hash.rb
ethn: good point

card/config/initializers/02_patches/kaminari.rb
```diff
@@ -12,6 +12,7 @@ def page_url_for page
           p = params_for(page)
           p.delete :controller
           p.delete :action
+          # TODO: make sure paging works with new related handling
```
ethn: I still need to check this, but since everything is inside a slot now, I think I can actually delete the following lines.

card/lib/card/error.rb
```diff
@@ -6,12 +6,15 @@ class Card
...
     cattr_accessor :current
 
-    class Oops < Error # carditect problem (rename!)
+    class BadContent < Error
     end
 
     class BadQuery < Error
     end
 
+    class BadAddress < Error
```
ethn: moved from decko because it is mentioned in card.

card/lib/card/error.rb
```diff
@@ -22,19 +25,19 @@ class UnknownCodename < NotFound
...
-      def initialize card
-        @card = card
-        super build_message
-      end
-
-      def build_message
-        if (msg = @card.errors[:permission_denied])
-          I18n.t :exception_for_card, scope: [:lib, :card, :error],
-                                      cardname: @card.name, message: msg
+      def initialize message
```
ethn: can now handled permission errors without cards.

card/lib/card/format/content.rb
```diff
@@ -1,20 +1,20 @@
...
         content_object.to_s
       end
 
-      def get_content_object content, opts
+      def get_content_object content, content_opts
         if content.is_a? Card::Content
           content
         else
-          Card::Content.new content, self, opts.delete(:content_opts)
+          Card::Content.new content, self, (content_opts || voo&.content_opts)
```
ethn: seems like a valid voo option for developers.

card/lib/card/format/context_names.rb
```diff
@@ -1,23 +1,25 @@
...
       def initial_context_names
-        @initial_context_names ||=
-          parent ? context_names_from_parent : context_names_from_params
+        @initial_context_names ||= relevant_context_names do
+          parent ? parent.context_names : context_names_from_params
+        end
       end
 
-      def context_names_from_parent
+      def relevant_context_names
```
ethn: this is an important step towards making it so that context names are handled entirely with the voo, because it gets rid of names that would needlessly kill caching.

having context names in the voo should eliminate some caching bugs where we see different variants of names than what we expect.

card/lib/card/format/nest/subformat.rb
```diff
@@ -27,6 +27,11 @@ def field_subformat field
           field = card.name.field(field) unless field.is_a?(Card)
           subformat field
         end
+
+        def inherit variable
```
ethn: this new api probably warrants a comment.  it means take the instance variable from me or the first ancestor that has it.

card/lib/card/view/options/voo_api.rb
```diff
@@ -65,13 +57,13 @@ def slot_options
...
-            ancestor.closest_live_option key
-          end
-        end
+        # def inherit key
+        #   if live_options.key? key
+        #     live_options[key]
+        #   elsif (ancestor = next_ancestor)
+        #     ancestor.inherit key
+        #   end
+        # end
```
ethn: I can still imagine a case where we might want this with legitimate options, but I refactored it away because it was only being used by foreign ones.

card/lib/card/view/options/voo_api.rb
```diff
@@ -138,26 +150,17 @@ def process_default_options
...
-        # non-standard options that are found in normalized_options
-        # @return [Hash] options Hash
-        def foreign_normalized_options
-          @foreign_normalize_options ||= foreign_options_in normalized_options
-        end
-
-        # non-standard options that are found in live_options
-        # @return [Hash] options Hash
-        def foreign_live_options
-          foreign_options_in live_options
```
ethn: farewell complexity.

card/mod/standard/set/rstar/rules_editor.rb
```diff
@@ -113,8 +113,8 @@ def link_to_open_rule
 
   def closed_rule_content rule_card
     return "" unless rule_card
-    nest rule_card, view: :closed_content,
-                    set_context: card.name.trunk_name
+    @set_context = card.name.trunk_name
+    nest rule_card, view: :closed_content
```
ethn: I proposed better api for this in the wikirate pr.

card/mod/standard/set/all/rich_html/content.rb
```diff
@@ -139,46 +138,27 @@ def current_set_card
...
   end
 
-  # subheader = with_name_context(card.name) do
-  #   subformat(rcard)._render_title(args)
-  # end
-
-  view :related, cache: :never do |args|
-    related_card, options = related_card_and_options args
+  # the related view nests a related card with a submenu
+  # the subcard is specified as an "item" card using the slot/voo api
```
ethn: comment above explains api change concept.  it's all about treating the related card as an item (which lets us get rid of a lot of old code)

card/mod/notifications/set/type/email_template/email_config.rb
```diff
@@ -1,61 +1,81 @@
...
-        default_from
-      end
-    config[:reply_to] ||= config[:from]
-  elsif config[:from].blank?
-    config[:from] = Card[Card::WagnBotID].account.email
+# @param [Card] context  the card in whose context all email fields will be interpreted
+# @param [Hash] fields override any templated field configurations with hash values
+# @param [Hash] opts options for rendering. unknown options become format options
+# @option opts [Card, String, Integer] :auth user identifier. render as this user
+def email_config context, fields={}, opts={}
```
ethn: this was actually the hardest api change.  all opts other than :auth are treated as format opts, which allows us to set instance variables in the format classes created.

Hmm, need to add that comment...

card/mod/notifications/set/type/email_template.rb
```diff
@@ -3,58 +3,41 @@ def clean_html?
...
+  mail = format.mail context, fields, opts
   mail.deliver
 rescue Net::SMTPError => exception
   errors.add :exception, exception.message
 end
 
 format do
-  def mail args={}
-    config = card.email_config args
+  def mail context=nil, fields={}, opts={}
```
ethn: more of our general move away from smushing all kinds of everything into "args"

PR #207, opened 2018-04-23
refactor nest handling into Nest class

ethn:


card/lib/card/env/location.rb
```diff
@@ -1,23 +1,9 @@
...
-        unless title.name?
-          Rails.logger.warn "Pass only Card::Name to page_path " \
-                            "(#{title} = #{title.class})"
-        end
-        format = opts[:format] ? ".#{opts.delete(:format)}" : ""
-        action = opts[:action] ? "#{opts.delete(:action)}/" : ""
-        query  = opts.present? ? "?#{opts.to_param}" : ""
-        card_path "#{action}#{title.to_name.url_key}#{format}#{query}"
-      end
-
```
ethn: got rid of this in favor of using format.path everywhere.

card/lib/card/format.rb
```diff
@@ -45,7 +45,7 @@ class Card
   #
   class Format
     include Card::Env::Location
-    include Nest
+    include Nesting
```
ethn: nesting has format methods.  Nest is the new class

card/lib/card/format/nest/fetch.rb
```diff
@@ -1,55 +1,58 @@
 class Card
   class Format
-    module Nest
+    class Nest
```
ethn: biggest change is introduction of this new class, which reduced the confusion from passing variables all around.

I think this class should get even simpler when modal handling is moved to Card::View, where it belongs.

card/lib/card/format/nesting.rb
```diff
@@ -0,0 +1,74 @@
...
+      include Mode
+
+      # @param cardish card mark
+      # @param options [Hash]
+      # @option options [Symbol] view
+      # @option options [Symbol, Array<Symbol>] :hide hide optional subviews
+      # @option options [Symbol, Array<Symbol>] :show show optional subviews
+      # @option options [String] :title
+      # @param override ??
+      def nest cardish, view_opts={}, format_opts={}
```
ethn: the newer, clearer nest API.

card/lib/card/format/nesting/subformat.rb
```diff
@@ -0,0 +1,52 @@
+class Card
+  class Format
+    module Nesting
+      # Formats can have subformats. Lineage can be retraced through "parent" format.
+      module Subformat
+        # note: while it is possible to have a subformat of a different class,
+        # the :format_class value takes precedence over :format.
+        def subformat subcard, opts={}
```
ethn: not sure why it didn't catch on to this renaming (nest to nesting), but the main change here is that now subformat takes opts.

card/lib/card/format/render.rb
```diff
@@ -74,8 +74,8 @@ def view_cache_setting view
 
       def stub_render cached_content
         result = expand_stubs cached_content do |stub_hash|
-          prepare_stub_nest(stub_hash) do |stub_card, mode, options, override|
-            with_nest_mode(mode) { nest stub_card, options, override }
+          prepare_stub_nest(stub_hash) do |stub_card, view_opts|
+            nest stub_card, view_opts, stub_hash[:format_opts]
```
ethn: simplified this by making stub representation very close to new nest api.

card/lib/card/query/join.rb
```diff
@@ -1,82 +1,125 @@
...
+      #
+      # all of the above can be set directly via opts using the keys with the same name.
+      #
+      #   Join.new side: "left", from_table: "cards"...
+      #
+      # The from and to fields can also be set via :from and :to keys.
+      # (see #interpret_from_and_to)
+      #
+      # You can generally use Symbols in place of Strings where applicable.
+      #
```
ethn: I think (hope) this makes more sense now.

card/spec/lib/card/format/render_spec.rb
```diff
@@ -8,6 +8,7 @@
 
   describe "view cache" do
     before { Cardio.config.view_cache = true }
+    after { Cardio.config.view_cache = false }
```
ethn: without this, view caching stays on for all tests that follow...  It's good to test caching but confusing to do it arbitrarily.

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -146,17 +146,17 @@ def process_edit_fields fields
   # then the subfield cards should be created on the new card not the existing
   # card that build the form
 
-
   def form
-    @form ||= begin
-      @form_root = true unless parent&.form_root
-      instantiate_builder(form_prefix, card, {})
-    end
+    @form ||= inherit(:form) || new_form
```
ethn: this means we don't have to pass @form through subformat explicitly.

card/spec/lib/card/format/render_spec.rb
xithan: ... and insane that this all we have to test the view cache system

ethn: y, I'll work on that.

ethn: (as part of the upcoming change when I switch over to the new opt-in caching.)

PR #208, opened 2018-04-25
UBOS-related changes:

ethn:
- made codename cache clear automatically when codename is changed
- cleaned up handling of schema dumping


card/mod/core/set/all/name_events.rb
```diff
@@ -0,0 +1,126 @@
...
+end
+
+def side_id_or_card old_name_in_way, sidecard, sidename
+  if !sidecard || old_name_in_way
+    add_subcard(sidename.s)
+  else
+    # if sidecard doesn't have an id, it's already part of this act
+    sidecard.id || sidecard
+  end
+end
```
xithan: I don't like that everything moves back in one big file. The original style guide rule (that we weakened because of too many violations) is not more than 100 lines. At least codenames and names could be separate files and I don't see why we can't keep the name validations event in a separate file.

card/lib/card/tasks/card/migrate.rake
```diff
@@ -39,20 +44,21 @@ namespace :card do
...
-      ActiveRecord::Base.dump_schema_after_migration = false
-      Cardio.schema_mode(:structure) do |paths|
-        ActiveRecord::Migrator.migrations_paths = paths
-        ActiveRecord::Migrator.migrate paths, version
-        Rake::Task["db:_dump"].invoke # write schema.rb
+      without_dumping do
+        Cardio.schema_mode(:structure) do |paths|
+          ActiveRecord::Migrator.migrations_paths = paths
+          ActiveRecord::Migrator.migrate paths, version
+        end
```
xithan: how do we update schema now?

xithan: Maybe we should add our own task for it. I don't think I will remember the  "db:_dump" task in the far future.

card/mod/core/set/all/name_events.rb
xithan: but how about subgrouping? The tedious validation stuff to name_events/name_validations.rb and keep the interesting stuff in name_events.rb or move it to something like name_events/name_changes.rb

xithan: I didn't notice all the other stuff in name.rb. That's indeed an improvement.

PR #209, opened 2018-04-26
skin fix

xithan:
and get migrations working for wikirate


card/mod/bootstrap/set/type/customized_skin.rb
```diff
@@ -85,7 +85,7 @@ def add_bootswatch_subfield
 end
 
 def theme_card
-  @theme_card ||= theme_codename && Card[theme_codename]
+  @theme_card ||= theme_codename && Card[theme_codename] || nil
```
ethn: usually we prefer `a ? b : c` over `a && b || c`, yes?

xithan: that's true

PR #213, opened 2018-04-29
reduce history tracking

ethn:


card/mod/account/set/right/token.rb
```diff
@@ -4,6 +4,10 @@
 
 card_accessor :expiration
 
+def history?
+  false
+end
```
ethn: in principle, this seems right; a token history seems like it could weaken security.

on the other hand, if we were having issues, this could help us debug them...

PR #214, opened 2018-04-30
skin migrations

xithan:
I vaguely remember that we needed the version.txt files for cldstr. Do we still need them for UBOS? 
I wasn't able to update the test seed data with that approach instead I added the migrations table to the seed data.


ethn:
I think we probably will want to revive the version.txt things before long, but given how much has changed, we can think of it as a rebuild. 

It's not strictly necessary for migrations to work; it's an optimization that helps avoid having to start up each deck to test whether migrations have to run. I would have to build all the framework around it that I built for cldstr, and I'd rather not tackle that right now.

ethn:
whether or not we have the textual stamping as an optimization, I definitely think we want the migrations table.  The stamp is to prevent it from running unnecessarily, but we have to have the table to track which migrations have run.

card/mod/bootstrap/db/migrate_core_cards/20180425174432_delete_deprecated_skin_cards.rb
```diff
@@ -0,0 +1,29 @@
+# -*- encoding : utf-8 -*-
+
+require_relative "lib/skin"
+
+class Skin
+  def delete_deprecated_skin_cards
```
ethn: Why do we need this in two migrations?

card/mod/bootstrap/db/migrate_core_cards/lib/skin.rb
```diff
@@ -15,9 +15,16 @@ def themes
       json = File.read File.join(vendor_path, "bootswatch", "docs", "api", "4.json")
       JSON.parse(json)["themes"].map { |theme| theme["name"] }
     end
```
ethn: I should make sure I understand how this lib works.  Is it ever auto loaded?

xithan: I think the root lib directory is in the autoload paths but this one shouldn't be there

PR #222, opened 2018-05-09
pointer reorganization

ethn:
includes change in handling of invalid items.


card/mod/notifications/set/type/notification_template.rb
```diff
@@ -29,5 +29,5 @@ def disappear?
 end
 
 def alert_class
-  contextual_class.present? ? contextual_class_card.item : :success
+  contextual_class.present? ? contextual_class_card.item_names.first : :success
```
xithan: why did you remove `item`?

card/mod/pointer/set/abstract/02_pointer/item_api.rb
```diff
@@ -0,0 +1,162 @@
...
+# @option args [String] :complete keyword to use in searching among items
+# @option args [True/False] :known_only if true, return only known cards
+# @option args [String] :type name of type to be used for unknown cards
+def item_cards args={}
+  return item_cards_search(args) if args[:complete]
+  return known_item_cards(args) if args[:known_only]
+  all_item_cards args
+end
+
+# ~~~~~~~~~~~~ ALTERING ITEMS ~~~~~~~~~~~~
```
xithan: if you need comments to organize a file you probably should split it in several files/modules

card/mod/notifications/set/type/notification_template.rb
ethn: It's inconsistent API.  we have #item_names, #item_id, and #item_cards.  To be consistent, #item would need to be If we want to support #item_name, #item_id, and #item_card, that would at least be consistent. But given that there's only one usage, it didn't seem worth it.

xithan: Often we only care about one item. We have a lot of `item_names.first` and `item_cards.first` calls in wikirate and whenever I write it I want something like `item`. I didn't it know it already existed.

card/mod/pointer/set/abstract/02_pointer/item_api.rb
ethn: not sure I agree. It would be fine to move the "altering" ones to another module. But, while I wanted to make a distinction between strongly and weakly supported API (and I think comments are appropriate for that), I definitely think it would be more confusing to have the "weakly" supported helper methods in a separate file.

fwiw, this file has well under a hundred lines of code now; it's just the documentation that makes it seem long.

card/mod/notifications/set/type/notification_template.rb
ethn: If it's #item_name, I guess I'm fine with it. #item breaks the pattern.

ethn: actually, what if it were like:  `def item_name index=0, args`

card/mod/pointer/spec/set/abstract/pointer/html_views_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    let(:pointer) do
+      pointer = Card["Sample Pointer"]
+      pointer << "A"
+      pointer
+    end
+
+    describe_views :core, :closed_content, :editor,
+                   :list, :autocomplete, :checkbox,
+                   :radio, :select, :multiselect do |view|
+      it "doesn't have errors" do
```
ethn: these 9 tests are new (and at least one – autocomplete – would have failed before)

ethn:
I was thinking this way of organizing modules might become a pattern:
```
./options_api.rb
./other_views.rb
./html_views_filter
./html_views_filter/filter_items.haml
./html_views_filter/filtered_list_input.haml
./html_views
./html_views/list_item.haml
./html_views/checkbox_input.haml
./html_views/radio_input.haml
./html_views/autocomplete_input.haml
./html_views/list_input.haml
./events.rb
./html_views_filter.rb
./html_views.rb
./item_api.rb
```
I like:
- seeing all the events in one place
- seeing all the views in one place when practical
- when that's not practical, seeing all the html views in one place
- when that's not practical, moving off a few related views (filter in this case)
- this naming convention (views, html_views, html_views_filter...)
- keeping API changes in the original file when practical (pointer.rb)
- moving API changes to specific `_api.rb` files when not

One nice thing about grouping views together is that it keeps the haml more organized.

card/mod/notifications/set/abstract/test_context.rb
```diff
@@ -7,7 +7,7 @@
   end
 
   def test_context_card
-    (tc = card.left.fetch(trait: :test_context)) && tc.item_cards.first
+    card.left.fetch(trait: :test_context)&.item_card
```
ethn: new API in use

card/mod/pointer/set/abstract/02_pointer/item_api.rb
```diff
@@ -0,0 +1,172 @@
...
+# @option args [String] :complete keyword to use in searching among items
+# @option args [True/False] :known_only if true, return only known cards
+# @option args [String] :type name of type to be used for unknown cards
+def item_cards args={}
+  return item_cards_search(args) if args[:complete]
+  return known_item_cards(args) if args[:known_only]
+  all_item_cards args
+end
+
+# #item_name, #item_id, and #item_card each return a single item, rather than an array.
```
ethn: here is the new API as discussed.

Now that it's here, I'm realizing it should be generalized. Searches should have item_name, too.

And, in fact, it occurs to me that the item args API should be as general as possible and mean the same thing for searches, pointers, and whatever else.  Some things are already quite good (offset, limit...).    Others not so much; right now `item_cards type: blah` restricts the results in searches and sets the type of new cards in pointers.  I'm thinking the more general API for the pointer use case should be `item_cards new: { type: blah }`. 

`content` has a similar problem, but I'm sure we can work around it.

It's also weird that "complete" is the only thing that triggers a search query. seems like you should be able to use any WQL.

More generally, it seems like we should harmonize things like "skip_virtual" (in fetch) and "known_only" (in item_cards)>
```
# for #fetch, #item_names, etc
new: false    # only real cards (equivalent to skip virtual)
new: :virtual # only return real cards if virtual (equivalent to known_only)
new: { ... }  # return new cards with these args
```

ethn: another thought: any time any of these plural item methods are called, we should be using MGET and loading up the cache. could reduce memcache calls quite a bit!

xithan:
maybe `html_views/filter.rb` and `html_views/filter/filter_items.haml` instead of `html_views_filter`?
or just the format as subdirectory: `html/filter.rb`

PR #224, opened 2018-05-12
reorganize and refactor follow/notify

ethn:
you can see the mod README rendered here: https://github.com/ethn/decko/tree/follow/card/mod/follow


card/mod/notify/set/type_plus_right/user/follow.rb
```diff
@@ -0,0 +1,72 @@
...
+# overrides pointer default
+def item_names _args={}
+  if (user = left)
+    Card.preference_names user.name, "follow"
+  else
+    []
+  end
+end
+
+# TODO: codename!
```
xithan: do it

PR #225, opened 2018-05-15
card_virtuals table

xithan:


card/mod/admin/set/self/admin.rb
```diff
@@ -30,7 +30,7 @@ def run_task_from_task_basket task
   abort :success
 end
 
-def not_alRAlowed task
+def not_allowed task
```
ethn: Won't this break all of our `not_alRAlowed` calls?  ;)

card/mod/admin/set/self/admin.rb
```diff
@@ -118,7 +118,7 @@ def solid_cache_count
   end
 
   def machine_cache_count
-    Card.search right: { codename: "machine_cache" }, return: "count"
+    Card::Virtual.where(right_id: Card.fetch_id(:machine_cache)).count
```
ethn: Card::MachineCacheID?

card/mod/machines/lib/javascript/decko.js.coffee
```diff
@@ -2,7 +2,7 @@ $.extend decko,
   # returns absolute path (starting with a slash)
   # if rawPath is relative (no slash), this adds relative root
```
ethn: need comment:
`# if rawPath is complete url, this returns the complete url`

card/mod/machines/set/abstract/machine.rb
```diff
@@ -144,7 +144,7 @@ def direct_machine_input? input_card
 
 def run_engine input_card
   return unless direct_machine_input? input_card
-  if (cached = fetch_cache_card(input_card))
+  if (cached = fetch_cache_card(input_card)) && cached.content.present?
```
ethn: maybe we make
```
def content?
  content.present?
end
```
...that way we can do `&.content?` here and elsewhere.

card/mod/machines/set/all/reset_machines.rb
```diff
@@ -1,12 +1,11 @@
...
   def reset_all_machines
     Auth.as_bot do
-      Card.search(
-        right: { codename: %w[in machine_cache machine_output] }
-      ).each do |card|
+      Card.search(right: { codename: "machine_output" }).each do |card|
         card.update_columns trash: true
         card.expire
       end
+      Card::Virtual.where(right_id: Card.fetch_id(:machine_cache)).delete_all
```
ethn: Card::MachineCacheID again?

card/mod/machines/set/all/reset_machines.rb
```diff
@@ -1,12 +1,11 @@
 module ClassMethods
   def reset_all_machines
     Auth.as_bot do
-      Card.search(
-        right: { codename: %w[in machine_cache machine_output] }
-      ).each do |card|
+      Card.search(right: { codename: "machine_output" }).each do |card|
```
ethn: not that it matters, but I think search takes a block, so the "each" isn't necessary.

card/mod/virtual/lib/card/virtual.rb
```diff
@@ -0,0 +1,77 @@
+# -*- encoding : utf-8 -*-
+
```
ethn: I had a moment where I thought maybe this name should be CardVirtualCache or CardCacheVirtual or some variant with "content" in it.

My main concern was that users who didn't understand virtual cards would assume that to be virtual, you had to be in this table.

I'm not falling in love with those, so I think we can go with this, but we need some docs here explaining what this is (and maybe pointing to some not-yet-existent documentation of virtual cards).

card/mod/virtual/set/abstract/virtual.rb
```diff
@@ -0,0 +1,45 @@
+# -*- encoding : utf-8 -*-
+
```
ethn: Can we make this name VirtualCache?

card/mod/machines/lib/javascript/decko_upload.js.coffee
```diff
@@ -5,7 +5,7 @@ $.extend decko,
 # only the file, type_id and attachment_card_name are needed
 # attachment_card_name is the original card name,
 # ex: card[subcards][+logo][image], card[file]
-    $(fileupload).bind 'fileuploadsubmit', (e,data) ->
+    $(fileupload).on 'fileuploadsubmit', (e,data) ->
```
ethn: wow, was that hard to find?

xithan: not really, jquery has a changelog for that

card/mod/virtual/spec/set/abstract/virtual_cache_spec.rb
```diff
@@ -0,0 +1,29 @@
+# -*- encoding : utf-8 -*-
+
+RSpec.describe Card::Set::Abstract::Virtual do
```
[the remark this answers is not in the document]
xithan: me too

PR #228, opened 2018-05-16
fix deprecated jquery code

xithan:


card/mod/bootstrap/db/migrate_core_cards/20180423143147_add_type_customized_bootswatch_skin.rb
```diff
@@ -4,9 +4,18 @@
...
+      if card.codename != "customized_bootswatch_skin"
+        delete_code_card :customized_bootswatch_skin
+      end
+      card.update_attributes! type_id: Card::CardtypeID,
+                              codename: "customized_bootswatch_skin"
+    else
+      ensure_card "Customized bootswatch skin",
+                  type_id: Card::CardtypeID,
+                  codename: "customized_bootswatch_skin"
+    end
```
ethn: I wonder if we can re-simplify this after we've run it on our own stuff.  I don't think this migration has been released yet, so only our own sites would be in a confused state.

PR #236, opened 2018-05-22
fixes for staging deployment

ethn:


card/lib/card/model/save_helper.rb
```diff
@@ -66,7 +66,7 @@ def ensure_card name_or_args, content_or_args=nil
           ensure_attributes card, args
           card
         else
-          Card.create! args.merge(name: name)
+          Card.create! args.reverse_merge(name: name)
```
ethn: this way you can ensure via codename.

xithan: I don't see a difference. At this point `args` has never a key `:name`. Moreover, if you give a name argument (like you do below in the migration) `ensure_card` will use the name to fetch the card, not the codename. So if the name of the existing codename card is different you will create a second card with a codename duplicate.

ethn: OK, I looked more closely at standardize_args, and you're right, this change isn't great.  But (A) it looks to me like you'd end up with `name: [Symbol]` here if you give a Symbol as the first arg.

In any case, we don't have duplicate codenames on staging, where this was run. Confusingly, we have no cards with a codename beginning with "customized" there.  Meanwhile, on the live server, we do have a card with that codename and the name "Customized Skin".

Maybe we can pair on some of these finish-line issues tomorrow.

xithan: A few lines above it says `args.delete(:name)`. It's impossible that there is a name here.

ethn: (that said, it's already in staging, so if it's possible to work from this commit, that would prevent a mess!)

card/mod/bootstrap/db/migrate_core_cards/20180423143147_add_type_customized_bootswatch_skin.rb
```diff
@@ -19,18 +19,10 @@ def up
...
-      end
-      card.update_attributes! type_id: Card::CardtypeID,
-                              codename: "customized_bootswatch_skin"
-    else
-      ensure_card "Customized bootswatch skin",
-                  type_id: Card::CardtypeID,
-                  codename: "customized_bootswatch_skin"
-    end
+    ensure_card :customized_bootswatch_skin,
+                name: "Customized bootswatch skin",
```
ethn: if we just do `ensure_card "Customized bootswatch skin",`..., then it works for most decks, yes?  Maybe we should just do that and then fix wikirate by hand?

PR #237, opened 2018-05-23
remove from tinymce menubar: file -> new document

xithan:


card/mod/tinymce_editor/db/migrate_cards/20171113233151_update_tinymce_config.rb
```diff
@@ -0,0 +1,22 @@
...
+      "relative_urls":false,
+      "extended_valid_elements":"a[name|href|target|title|onclick],img[class|src|border=0|alt|title|hspace|vspace|width|height|align|onmouseover|onmouseout|name],hr[class|width|size|noshade],font[face|size|color|style],span[class|align|style]"
+     }
+  JSON
+
+  def up
+    ensure_card :tiny_mce,
+                content: TINYMCE_CONF
+  end
+end
```
ethn: Are we sure we want to overwrite everyone's tinyMCE conf?  Can we only update pristine?

ethn:
I do think updating this config for folks who haven't customized theirs would be a good idea.  I assume we still have that "pristine" API?

PR #239, opened 2018-05-23
abstracted query "tying"

ethn:
It's increasingly evident that no single strategy ("join", "where exists", etc) will work everywhere on a large site; there has to be some capacity to optimize card queries.

This PR is a big step in that direction, because it makes it so that you can easily change strategies.

The API could use some further thought, and I need to spend some time organizing Query tests, but when I tweak the DEFAULT_FASTEN value by hand, all query_spec tests for all three values. 

Note that we could easily extend this same approach and add an option to break a single query into multiple ones.


card/lib/card/query/abstract_query/tie.rb
```diff
@@ -0,0 +1,78 @@
...
+      #
+      # In concept, here's how the different strategies would tie table A to table B
+      # in SQL assuming A.id = B.a_id
+      #
+      # - :join  ...  FROM A JOIN B ON A.id = B.a_id
+      # - :exist ...  FROM A WHERE EXISTS (SELECT * FROM B WHERE A.id = B.a_id ...)
+      # - :in    ...  FROM A WHERE A.id IN (SELECT B.a_id FROM B WHERE ...)
+      #
+      # The different strategies will return the same values but the relative speed is
+      # context dependent.
```
ethn: This explains the tying strategies.

ethn:
For now, the default is set back to :join, so it's basically back to the old style queries for now, and then we can cherry pick slow ones to optimize.

card/lib/card/query/abstract_query/tie.rb
xithan: Where are the tests that prove that it really works like that?

ethn: coming.

card/spec/lib/card/query/card_query/special_attributes_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      it "it can handle *" do
+        @query = { match: "*all" }
+        is_expected.to include("*all")
+      end
+    end
+  end
+
+  describe "found_by" do
+    before do
+      Card::Auth.as_bot
```
xithan: I think doing it like this makes it persistent through all tests executed afterwards
Better use the `as_bot: true` option.

card/spec/lib/card/query/card_query/special_attributes_spec.rb
```diff
@@ -0,0 +1,76 @@
+require_relative "../query_spec_helper"
+RSpec.describe Card::Query::CardQuery::SpecialAttributes do
+  include QuerySpecHelper
+
+  # TODO: add specs for: #name_match, #complete, #junction_complete
+
+  describe "match" do
+    it "reachs content and name via shortcut" do
+      @query = { match: "two" }
+      is_expected.to eq(cards_matching_two)
```
xithan: The style guide (which is disabled for theses files) dislikes instances variables and he has a point.
It obfuscates what the test is doing.

card/lib/card/query/abstract_query/tie.rb
ethn: (they are there now)

card/spec/lib/card/query/card_query/special_attributes_spec.rb
ethn: I agree. This PR didn't introduce that, though. Can we go ahead and merge this?  I want to tackle the actual performance problems before getting into more spec clean up.

card/spec/lib/card/query/card_query/special_attributes_spec.rb
ethn: for now I just used the as_bot block.  That way the queries are run with standard permissions.

card/spec/lib/card/query/card_query/special_attributes_spec.rb
ethn: ok, I took care of it.  should be ready now.

PR #240, opened 2018-05-25
fixed related view bug

ethn:


card/mod/standard/set/all/rich_html/related.rb
```diff
@@ -0,0 +1,31 @@
...
+    voo.show :toolbar, :menu, :help
+    frame do
+      voo.hide :header, :toggle
+      nest @related_card, related_options(options)
+    end
+  end
+
+  def related_item
+    options = voo.items
+    voo.items = {}
```
ethn: this line (`voo.items = {}`) is the difference.  we don't want the configuration of a specific related view to get stored in the slot, because it interferes with all the other related views.

PR #243, opened 2018-05-28
Fastacts

ethn:
optimize the all_viewable query.

(will be very easy to see once the query request is merged)


xithan:
... and the conflicts are resolved

card/mod/history/lib/card/action.rb
```diff
@@ -25,6 +25,7 @@ class Action < ApplicationRecord
     extend Card::Action::Admin
 
     belongs_to :act, foreign_key: :card_act_id, inverse_of: :ar_actions
+    belongs_to :ar_card, foreign_key: :card_id, inverse_of: :actions, class_name: "Card"
```
ethn: Neither of likes this nomenclature.

I would propose the following (but I'm not sure if it works):
1. for acts, actions, changes, and references, we use the full name (eg `card_acts`) to refer to the active record relations and the short name (eg `acts`) to refer to the caching version.
2. for cards, we can't do this, but I think it's pretty important that, eg, `@action.card` use the cache.  if it's possible to use both `belongs_to :card` and `def card` harmoniously (eg, so that we can do `joins(card_actions: :card) above`), that would be great.  But I'm not sure it will work.  If not `ar_card`, what is another option?  `card_model`?  `card_relation`?

card/mod/history/lib/card/act.rb
```diff
@@ -42,18 +42,17 @@ def delete_actionless
...
       # all acts with actions that current user has permission to view
-      # @return [Array of Actions]
+      # @return [ActiveRecord Relation]
       def all_viewable action_where=nil
-        viewable_actions = Action.all_viewable.where "card_acts.id = card_act_id"
-        viewable_actions = viewable_actions.where(action_where) if action_where
-        where("EXISTS (#{viewable_actions.to_sql})").where.not "card_id" => nil
+        relation = joins(ar_actions: :ar_card)
+        relation = relation.where(action_where) if action_where
+        relation.where(Query::CardQuery.viewable_sql).where.not(card_id: nil).distinct
```
ethn: What's important about this PR is that for this query, the `JOIN` strategy is much faster than `WHERE EXISTS`.

I contemplated trying to build this with ActQuery rather than ActiveRecord relations.  I'm sure it would be doable with a little work. But this is a fairly urgent matter, so I think we need to get this out before having any big discussions about how the two querying systems fit together.

PR #246, opened 2018-05-30
buncha unrelated bugfixes

ethn:


card/mod/core/chunk/reference.rb
```diff
@@ -8,12 +8,15 @@ class Reference < Abstract
...
         def referee_name
           return if name.nil?
-          @referee_name ||= begin
-            rendered_name = render_obj(name)
-            ref_card = fetch_referee_card rendered_name
-            ref_card ? ref_card.name : rendered_name.to_name
-          end
+          @referee_name ||= referee_name_from_rendered(render_obj(name))
           @referee_name = @referee_name.absolute(card.name).to_name
+        rescue Card::Error::NotFound
```
ethn: fixes voting bug.  this continues the (debatable) strategy of allowing pointers to contain illegal references, or at least not breaking weirdly when they do. Our current situation on wikirate is that many pointers have id references to cards that have since been deleted. Arguably, if we're using ids, we have to enforce referential integrity (ie, you can't delete a card without deleting all references to its id).  names do not have this problem...

anyway, I don't want to get stuck on voting stuff when we're about to replace it with bookmarks, so this is a quick fix.

card/mod/machines/lib/javascript/decko_editor.js.coffee
```diff
@@ -47,12 +47,10 @@ doubleClickApplies = (el) ->
   # false if there is a card-editor open inside slot
   slot.data 'cardId'
 
-
-
 triggerDoubleClickEditingOn = (el)->
   slot = el.slot()
   slot.addClass 'slotter'
-  slot.attr 'href', decko.path('~' + slot.data('cardId') + '?view=edit')
+  slot[0].href = decko.path('~' + slot.data('cardId') + '?view=edit')
```
ethn: double clicking is busted without this, thanks to the js update.

card/mod/standard/set/all/rich_html/editing.rb
```diff
@@ -139,7 +139,7 @@ def edit_type_buttons
   view :edit_structure, cache: :never do
     return unless card.structure
     voo.show :toolbar
-    render_related items: { view: :edit, nest_name: card.structure_card.name }
+    render_related items: { view: :edit, nest_name: card.structure_rule_card.name }
```
ethn: edit > structure busted without this.  I'll try to write quick tests for this and the next one, but I mostly just fixed it so that I could get something done.  (it's another one where we're planning to replace the UI, but maybe not quite as soon as with bookmarking)

ethn: (tests written)

PR #248, opened 2018-05-31
PT - #156623824 (use BT4 color variables) 

srivig:


card/mod/machines/lib/stylesheets/style_cards.scss
```diff
@@ -17,6 +17,31 @@ If you don't want certain rules, you may:
...
+$d-red-2: #966;
+$d-blue: #069;
+$d-cyan: #8cf;
+
+/*-- yellows --*/
+$d-yellow-1: #ffd;
+$d-yellow-2: #ffc;
+$d-yellow-3: #ffa;
+$d-yellow-4: #cc7;
+
```
xithan: In the current implementation this file can't use bootstrap variables but I think that's wrong. Although we want to allow people to get rid of bootstrap styles if they really want to that doesn't mean that we have to support a possibility to get rid of bootstrap variables.
I think we shouldn't duplicate any variables that are already in bootstrap. Do we have to add additional variables like these different yellow colors or can we can rid of them by using other bootstrap color variables?  The comments in the code indicate that variables like `$warning` or `$info` might be a good replacement.

xithan:
the PR that makes bootstrap variables availabe in `style_cards.scss`: #250

PR #250, opened 2018-06-05
allow empty json cards

xithan:


card/mod/bootstrap/set/abstract/bootswatch_theme.rb
```diff
@@ -45,13 +45,13 @@
...
 PRE_VARIABLES_CARD_NAMES = %i[
   style_jquery_ui_smoothness
-  style_cards
   style_right_sidebar
   bootstrap_functions
 ].freeze
 
 POST_VARIABLES_CARD_NAMES = %i[
   bootstrap_core
+  style_cards
```
ethn: how bad is it that this line isn't in 0.5?

card/mod/basic_types/set/type/json.rb
```diff
@@ -1,5 +1,5 @@
 event :validate_json, :validate, on: :save, changed: :content do
-  check_json_syntax
+  check_json_syntax if content.present?
```
ethn: ok for now.  wondering if we shouldn't have a default rule setting content to {}?

card/mod/bootstrap/set/abstract/bootswatch_theme.rb
xithan: no problem. we need that only for #248

card/mod/basic_types/set/type/json.rb
xithan: I agree in principal but "" is only a bad query, "{}" searches for all cards. That could cause a lot of performance trouble.

PR #251, opened 2018-06-06
handling bad slots and bad file configurations

ethn:


card/mod/machines/lib/javascript/decko_slot.js.coffee
```diff
@@ -27,12 +27,14 @@ $.extend decko,
...
   slot: (status="success") ->
-    if @data("slot-#{status}-selector")
-      @findSlot(@data("slot-#{status}-selector"))
-    else if @data("slot-selector")
-      @findSlot(@data("slot-selector"))
-    else
-      @closest '.card-slot'
+    @selectSlot("slot-#{status}-selector") ||
+      @selectSlot("slot-selector") ||
+      @closest(".card-slot")
```
ethn: in addition to the increased handsomeness, there is a functional change: if the slot is not found by the selector, it moves on to the next option.

Eg, +formula cards have a configured slot selector so that the entire metric card is updated.  This means that if you edit the +formula card on its own (and there is no metric slot) it can't find the slot it wants.  In the old system it would just quietly fail to update anything.  With this change it falls back to updating its own slot.

PR #252, opened 2018-06-06
permless views

ethn:
Sites with restricted permissions have lots of problems in head tags without these fixes.


ethn:
This is probably significant enough to rush a 0.5.1 release out.

PR #260, opened 2018-06-22
cascade name changes

xithan:
use the subcard api to traverse the tree of 
descendants that are affected by a name change


card/mod/core/set/all/rename.rb
```diff
@@ -14,22 +14,37 @@ def suspend_name name
...
+    Rails.logger.info "cascading name: #{child.name}"
+    newname = child.name.swap name_before_last_save, name
+    # check_for_conflict de.name, newname
+    Card.expire child.name # old name
+    attach_subcard child.name,
+                   superleft: self,
+                   name: newname, update_referers: update_referers
+
+
+    # Card.where(id: de.id).update_all name: newname.to_s, key: newname.key
```
ethn: any reason to keep these comments?

card/mod/core/set/all/rename.rb
```diff
@@ -14,22 +14,37 @@ def suspend_name name
...
-    check_for_conflict de.name, newname
-    Card.expire de.name # old name
-    Card.where(id: de.id).update_all name: newname.to_s, key: newname.key
-    de.update_referers = update_referers
-    de.refresh_references_in
-    Card.expire newname
+    Rails.logger.info "cascading name: #{child.name}"
+    newname = child.name.swap name_before_last_save, name
+    # check_for_conflict de.name, newname
+    Card.expire child.name # old name
```
ethn: if we leave this, can we leave a comment that says we're not sure it's needed so we don't assume it was added with good reason?

PR #269, opened 2018-07-19
rename and reference fixes

ethn:


card/mod/core/spec/set/all/rename_spec.rb
```diff
@@ -159,6 +159,12 @@ def assert_rename card, new_name
       expect_content_of("blue linker 2").to eq "[[Red]]"
     end
 
+    it "handles link to and nest of same card" do
```
ethn: this test fails without the new "uniq" handling in references.  This is why I couldn't rename "Inclusion Syntax" on decko.org.

card/mod/core/set/all/references.rb
```diff
@@ -9,24 +9,30 @@
...
 def referers
-  references_in.map(&:referer_id).map(&Card.method(:fetch)).compact
+  referer_cards_from_references references_in
 end
 
 # cards that include self
-def includers
-  refs = references_in.where(ref_type: "I")
-  refs.map(&:referer_id).map(&Card.method(:fetch)).compact
+def nesters
```
ethn: main changes:
1. "include" to "nest" in API
2. these card methods return uniq cards.

card/mod/core/spec/set/all/references_spec.rb
```diff
@@ -1,9 +1,44 @@
...
   it "replaces references should work on nests inside links" do
     card = Card.create! name: "ref test", content: "[[test_card|test{{test}}]]"
     assert_equal "[[test_card|test{{best}}]]",
                  card.replace_reference_syntax("test", "best")
   end
+
+  describe "#referers" do
+    it "returns all cards that refer to card" do
+      expect(Card["Blue"].referers.map(&:name).sort)
+        .to eq(["blue includer 1", "blue includer 2", "blue linker 1", "blue linker 2"])
```
xithan: if the order doesn't matter use `contain_exactly` matcher (instead of `sort`)

card/mod/core/spec/set/all/rename_spec.rb
```diff
@@ -159,6 +159,12 @@ def assert_rename card, new_name
...
+      update "blue linker 1", content: "[[Blue]] is {{Blue|name}}"
+      update "Blue", name: "Red", update_referers: true
+      expect_content_of("blue linker 1").to eq "[[Red]] is {{Red|name}}"
```
xithan: we have better api for that: `expect_card("...").to have_content "..."

PR #270, opened 2018-07-23
only run "changed" events on update

ethn:
There was an issue in wikirate where a `changed: :content` condition was failing on create, because db_content was never set.  I don't think this is a distinction we want to try to explain.  It's much easier to say that creation automatically changes all fields by definition.

### DIFF
--- a/card/mod/standard/set/type/session.rb
+++ b/card/mod/standard/set/type/session.rb
@@ -23,7 +23,7 @@ def content
 end
 
 def content= val
-  Env.session[key] = val
+  self.db_content = Env.session[key] = val
 end
 
 format :html do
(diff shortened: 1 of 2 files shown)


card/mod/standard/set/type/session.rb
```diff
@@ -23,7 +23,7 @@ def content
 end
 
 def content= val
-  Env.session[key] = val
+  self.db_content = Env.session[key] = val
```
ethn: Fixes session-related source issue in wikirate cukes.  Bug appears to have been introduced here: https://github.com/wikirate/wikirate/compare/43795ab...725bc17

I'm not sure why this became necessary, but the most likely candidate so far is  `attribute_is_changing?`

Without this, db_content is often nil, so store_in_session just sticks a nil value in the session.

PR #272, opened 2018-07-24
update reset password to use new trigger api

ethn:
also generally reordered account and signin pages.


card/mod/account/set/right/account.rb
```diff
@@ -8,6 +8,70 @@
...
+event :send_account_verification_email, :integrate,
+      on: :create, when: proc { |c| c.token.present? } do
+  Card[:verification_email].deliver self, to: email
+end
+
+# ON UPDATE
+
+# reset password emails contain a link to update the +*account card
+# and trigger this event
+event :reset_password, :prepare_to_validate, on: :update, trigger: :required do
```
ethn: this uses new API.

card/mod/account/set/right/account.rb
```diff
@@ -39,26 +116,60 @@ def refreshed_token
...
+  act.actions_affecting(act.card).each do |action|
+    return true if action.card.ok? :read
+  end
+  false
+end
+
 format do
   view :verify_url, cache: :never do
-    card_url path({ mark: card.name.left }.merge(token_path_opts))
+    path token_path_opts.merge(mark: card.name.left)
```
ethn: card_url not necessary. #path only produces relative urls in html format.

card/mod/account/set/self/signin.rb
```diff
@@ -1,8 +1,95 @@
...
+end
+
+event :signout, :validate, on: :delete do
+  Auth.signin nil
+  abort :success
+end
+
+# triggered by clicking "Reset my Password", this sends out the verification password
+# and aborts (does not sign in)
+event :send_reset_password_token, before: :signin, on: :update, trigger: :required do
```
ethn: new API

card/mod/account/set/right/account.rb
```diff
@@ -8,6 +8,70 @@
 card_accessor :status
 card_accessor :token
 
+#### ON CREATE
+
```
ethn: this is mostly reordering to what I'm promoting as our default order (events on top, then card methods, then  formats (within which we go views then methods)

card/mod/standard/set/all/rich_html/toolbar.rb
```diff
@@ -212,12 +212,17 @@ def toolbar_pin_button
 
   view :delete_button do
     confirm = "Are you sure you want to delete #{safe_name}?"
-    success = main? ? "REDIRECT: *previous" : "TEXT: #{card.name} deleted"
+    success = main? ? "REDIRECT: *previous" : { view: :just_deleted }
```
ethn: I removed support for this. Seems inconsistent and maybe insecure.

decko/rails/controllers/card_controller.rb
```diff
@@ -49,37 +49,43 @@ def asset
...
+    handle_errors do
+      @card = Card.controller_fetch params
+      raise Card::Error::NotFound unless card
+      record_as_main
+    end
   end
 
   def load_action
-    @card.select_action_by_params params
+    handle_errors do
```
ethn: I changed this logic a bit.  we don't need to call load_action when we're loading cards for create/update/delete, do we?

decko/rails/controllers/card_controller.rb
```diff
@@ -96,27 +102,27 @@ def handle
...
+
+  # successful create, update, or delete action
   def render_success
-    success = Card::Env.success
-    success.name_context = @card.name
-    if !Card::Env.ajax? || success.hard_redirect?
-      card_redirect success.to_url
-    elsif success.target.is_a? String
-      render plain: success.target
+    success = Card::Env.success.in_context card.name
```
ethn: this makes the soft vs hard redirect much more intelligible, imo

decko/lib/decko/rest_spec_helper.rb
```diff
@@ -0,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+
+module Decko
+  # For use in REST API specs
+  module RestSpecHelper
+    def self.describe_api &block
+      RSpec.describe CardController, type: :controller do
+        routes { Decko::Engine.routes }
+        include Capybara::DSL
+        instance_eval(&block)
```
xithan: Can't we do this with a shared context that includes `RSpec::Rails::ControllerExampleGroup`?

ethn: I could believe it, but that will have to wait until after vacation :)

PR #277, opened 2018-08-09
specifications for view

dcastroeyss:


card/mod/admin/set/self/admin_info.rb
```diff
@@ -26,10 +26,10 @@ def warning_list warnings
   end
 
   def warning_list_with_auto_scope warnings
-      # 'ADMINISTRATOR WARNING'
-      admin_warn = tx(:admin_warn)
-      "<h5>#{admin_warn}</h5>" + warnings.join("\n")
-    end
+    # 'ADMINISTRATOR WARNING'
+    admin_warn = tx(:admin_warn)
```
ethn: is this supposed to be "tr"?

card/mod/admin/spec/set/self/admin_info_spec.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+
+describe Card::Set::Self::AdminInfo do
+  # SPECSTUB
```
ethn: SPECSTUB means there aren't any specs yet.  So we can delete these.

card/mod/admin/set/self/admin_info.rb
dcastroeyss: it looks like it is

card/mod/admin/spec/set/self/admin_info_spec.rb
dcastroeyss: ok

card/mod/admin/spec/set/self/admin_info_spec.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+
+describe Card::Set::Self::AdminInfo do
```
xithan: `describe` on the top level still works but it's a deprecated pattern that RSpec pollutes the global namespace like that. The better way is `RSpec.describe`.
IDE likes Rubymine also expect it that way.

fyi you can generate these files with `bundle exec decko generate card:set admin self admin_info --spec-only`

dcastroeyss: `RSpec.describe Card::Set::Self::AdminInfo do....`
Yes?

ethn: yes.

ethn: Yes.

card/mod/admin/spec/set/self/trash_spec.rb
```diff
@@ -0,0 +1,8 @@
+# -*- encoding : utf-8 -*-
+
+describe Card::Set::Self::Trash do
+  # SPECSTUB
```
ethn: (there are still a few of these)

PR #280, opened 2018-08-15
json item fixes

ethn:


card/mod/basic_formats/set/all/json.rb
```diff
@@ -97,9 +97,7 @@ def name_search query_attribute: :complete, limit: AUTOCOMPLETE_LIMIT
   end
 
   view :items, cache: :never do
-    item_cards.map do |i_card|
-      nest i_card
-    end
+    listing item_cards
```
ethn: `?item=X` not working without this

card/mod/basic_formats/set/all/json.rb
```diff
@@ -139,6 +137,12 @@ def name_search query_attribute: :complete, limit: AUTOCOMPLETE_LIMIT
     }
   end
 
+  view :export_item do
+    item = { name: card.name, type: card.type_name }
+    item[:codename] = card.codename if card.codename
+    item
```
ethn: not sure this is the best name, but we need a view that gives us only importable attributes (not id, url, etc)

PR #281, opened 2018-08-20
reorganized history code

ethn:
events together, views together, other stuff more thematically aligned.


card/mod/history/set/all/actions.rb
```diff
@@ -1,185 +0,0 @@
...
-  if args[:rev]
-    nth_action args[:rev]
-  elsif args[:rev_id].is_a?(Integer) || args[:rev_id] =~ /^\d+$/
-    if (action = Action.fetch(args[:rev_id])) && action.card_id == id
-      action
-    end
-  # revision id is probably a mod (e.g. if you request
-  # files/:logo/standard.png)
-  elsif args[:rev_id]
-    last_action
```
ethn: this is one of the few bits of logic I intentionally changed.  Rather than explicitly setting the last action for a revision_id we can't parse, I just leave it unset. If there's a good reason to do this, we should document it.

ethn:
I'm holding off on making deeper changes until the reorganization is approved. (so it's easier to see the real changes)

ethn:
see https://www.pivotaltracker.com/story/show/160142676 
and https://www.pivotaltracker.com/story/show/160149048

ethn:
to me, the point was to make it so that all the functionality was in All::History, as opposed to modules whose make sense in the context of history but not outside of that context.  Probably more taste than anything, but we should settle on a convention.

xithan:
So the problem is that the mod is not part of the module structure

PR #282, opened 2018-08-24
matching and other wikirate-related fixes.

ethn:


card/lib/card/query.rb
```diff
@@ -53,8 +53,8 @@ module Query
                           refer_to referred_to_by
                           member_of member
 
-                          found_by not sort match name_match complete
-                          junction_complete extension_type],
```
ethn: `name_match` and `junction_complete` seem like hacky add-ons.  We can already achieve `name_match` with `{ name: ["match", X] }`.  (it needs improvement, but we don't want to add new attributes to avoid fixing things).

junction_complete is a weirder (metric-specific) need but was being used as our default complete pattern.  It slows down normal completions and really isn't necessary, especially with my name_match simplifications.

card/lib/card/query/card_query/attribute_helper.rb
```diff
@@ -22,24 +22,6 @@ def junction_val val, side
...
-          %(#{field} #{cxn.match quote("[[:<:]]#{val}[[:>:]]")})
-        end
-
-        def name_like patterns, extra_cond=""
-          likes =
-            Array(patterns).map do |pat|
-              "lower(#{table_alias}.name) LIKE lower(#{quote pat})"
-            end
-          add_condition "#{or_join(likes)} #{extra_cond}"
-        end
```
ethn: I think it's pretty odd that we were using or logic here.

card/lib/card/query/card_query/match_attributes.rb
```diff
@@ -0,0 +1,63 @@
...
+          found_by_cards(val).compact.each do |card|
+            subquery found_by_statement(card).merge(fasten: :direct, context: card.name)
+          end
+        end
+
+        # Implements the match attribute that matches always against content and name.
+        # Currently that's different from the match operator that can be restricted to
+        # names or content.
+        # Example: { match: "name or content" } vs. { name: ["match", "a name"] }
+        # TODO: unify handling for both using full text indexing
```
ethn: this TODO is the most important comment.  MySQL full text indexing has come a long way since we started working on this.  I'm hoping we can build a general solution for postgres and mysql and use that in all our matching.

card/mod/core/set/all/trash.rb
```diff
@@ -112,11 +112,15 @@ def db_attributes
...
-    add_subcard child
+    delete_as_subcard child
     # next if child.valid?
     # child.errors.each do |field, message|
     #   errors.add field, "can't delete #{child.name}: #{message}"
     # end
   end
 end
+
+def delete_as_subcard subcard
```
ethn: I use this in wikirate multiple times.

card/mod/machines/lib/javascript/decko.js.coffee
```diff
@@ -6,7 +6,7 @@ $.extend decko,
     if rawPath.match /^\/|:\/\//
       rawPath
     else
-      decko.rootUrl + '/' + rawPath
+      decko.rootUrl + rawPath
```
ethn: this was always producing double slashes.

card/mod/pointer/lib/javascript/script_pointer_config.js.coffee
```diff
@@ -37,7 +37,7 @@ $.extend decko,
   initAutoCardPlete: (input) ->
     optionsCard = input.data 'options-card'
     return unless !!optionsCard
-    path = optionsCard + '.json?view=name_complete'
+    path = optionsCard + '.json?view=name_match'
```
ethn: use matching, rather than just completion, in pointers.

card/lib/card/query/card_query/match_attributes.rb
```diff
@@ -0,0 +1,63 @@
...
+          field_match "#{table_alias}.db_content", val, cxn
+        end
+
+        def field_match field, val, cxn
+          %(#{field} #{cxn.match quote("[[:<:]]#{val}[[:>:]]")})
+        end
+
+        # TODO: move sql to SqlStatement
+        def name_like pattern, no_junction
+          conds = ["#{table_alias}.key LIKE #{quote pattern}"]
```
ethn: all these changes were inspired by the fact that the lower(name) queries were taking like 3s each on wikirate. That's been causing a lot of problems with the navbox and makes the autocomplete almost unusable.  Doing it with key instead cuts it down to a small part of a second.

card/mod/account/set/self/account_links.rb
```diff
@@ -1,7 +1,15 @@
...
+    status_class = Auth.signed_in? ? "logged-in" : "logged-out"
+    wrap_with :span, id: "logging", class: status_class do
+      item_links.join " "
+    end
+  end
+
   def item_links _args=nil
-    %i[my_card invite sign_out sign_up sign_in].map do |link_view|
+    # removed invite for now
+    %i[my_card sign_out sign_up sign_in].map do |link_view|
```
ethn: by request from folks at wikirate.  I think it's valid.

card/mod/standard/set/all/rich_html/content.rb
```diff
@@ -147,7 +147,7 @@ def current_set_card
   def rule_based_help
     return "" unless (rule_card = card.help_rule_card)
     with_nest_mode :normal do
-      process_content _render_raw(structure: rule_card.name), chunk_list: :references
+      process_content rule_card.content, chunk_list: :references
```
ethn: fixes help text in places where raw view is overridden.

card/spec/lib/card/query/card_query/sorting_spec.rb
```diff
@@ -22,7 +22,10 @@
   end
 
   it "plays nice with match" do
-    expect(run_query(match: "Z", type: "Basic", sort: "content"))
+    expect(run_query(match: "Z",
+                     not: { match: "Prose" },
```
ethn: new matching was matching some prosemirrorz thing...

card/mod/search/set/self/search.rb
```diff
@@ -1,5 +1,3 @@
-include_set Abstract::SearchParams
```
xithan: Are you sure that you want to remove that?
I see methods from that abstract set called in this file, e.g. `query_params`

xithan: Are you sure that you want to remove that?
I see methods from that abstract set called in this file, e.g. query_params

ethn: :search is a Search card.  Search has Abstract::Search.  Abstract::Search has Abstract::SearchParams.  So we're good. (it's there, I checked.)

card/mod/basic_formats/set/all/json.rb
```diff
@@ -29,30 +27,6 @@ def show view, args
...
   end
 
-  # TODO: design better autocomplete API
-  # view :name_complete, cache: :never do
-  #   name_search
-  # end
-
-  view :name_complete, cache: :never do
-    name_search query_attribute: :junction_complete
-  end
```
ethn: these views depend on `card.search`, which not all cards can access.  That's why I moved them into abstract/search.

PR #283, opened 2018-08-28
new layout handling and bridge

xithan:
the first step towards the "bridge"
working but needs clean up before merge


card/lib/card/content/chunk.rb
```diff
@@ -39,10 +39,11 @@ def register_list key, list
               next unless (p_map = prefix_map_by_chunkname[chunkname])
```
xithan: refactoring and added helpful error message if the chunk list key doesn't exist

card/db/migrate_core_cards/20180905061534_migrate_layouts.rb
```diff
@@ -0,0 +1,56 @@
+# -*- encoding : utf-8 -*-
+
+class MigrateLayouts < Card::Migration::Core
+  def up
+    Card.search type: "Layout" do |card|
+      update_layout_card card
+      create_head_rules card
```
xithan: Layouts now start with body tag. The surrounding html stuff is added by `page_wrapper` in the html format. The head is defined by `*head` rules.

card/lib/card/set/format/wrapper.rb
```diff
@@ -0,0 +1,89 @@
...
+        # Example
+        #   view :big_mac, wrap: { burger: { bun: "sesame" }, paper: { color: :red } }
+        #   view :cheese_burger, wrap: [:burger, paper: { color: :yellow }]
+        #
+        # If you want to define a wrapper that wraps only with a single html tag
+        # then use the following syntax:
+        #   wrapper :burger, :div, class: "medium"
+        #
+        #   wrap_with_burger "meat"  # => "<div class='medium'>meat</div>"
+        def wrapper wrapper_name, *args, &wrap_block
```
xithan: new API. see specs for more examples

card/VERSION
```diff
@@ -1 +1 @@
-0.6.7
+0.6.8
```
ethn: This should be 0.7.0

card/lib/card/set/format/abstract_format.rb
```diff
@@ -17,6 +20,7 @@ def before view, &block
         end
 
         def view view, *args, &block
+          # binding.pry
```
ethn: remove?

card/mod/edit/set/all/bridge/tab_views.rb
```diff
@@ -0,0 +1,73 @@
...
+    class_up "d0-card-body",  "history-slot"
+    voo.hide :act_legend
+    acts_bridge_layout card.history_acts
+  end
+
+  view :related_tab do
+    bridge_pills bridge_pill_items(RELATED_ITEMS, "Related")
+  end
+
+  view :rules_tab do
```
ethn: need `unknown_ok` here, right?  (Because I think we want to be able to see rules for new cards)

card/mod/edit/set/all/bridge/tab_visibility.rb
```diff
@@ -0,0 +1,51 @@
...
+  def show_history_tab?
+    card.real?
+  end
+
+  def show_related_tab?
+    card.real?
+  end
+
+  def show_rules_tab?
+    true
```
ethn: I suspect some sharks are going to want to be able to hide this tab from butterflies/eagles.  But this is a very clean starting place.

card/mod/edit/set/all/form.rb
```diff
@@ -177,7 +180,7 @@ def explicit_form_prefix
   end
 
   def form_context
-    (form_root? || !form_root) ? self : parent
+    form_root? || !form_root ? self : parent
```
ethn: parens around last three would be clearer

card/mod/edit/set/all/toolbar.rb
```diff
@@ -13,9 +13,7 @@ def toolbar_pinned?
   end
 
   view :toolbar, cache: :never do
-    tool_navbar do
-      [toolbar_split_buttons, toolbar_right_buttons]
-    end
+    ""
```
ethn: buh bye.

ethn: can we delete a bunch of this?

card/mod/follow/set/all/follow/start_follow_link.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class StartFollowLink < FollowLink
+  def initialize format
+    @rule_content = "*always"
+    @link_text = "follow"
+    @action = "send"
+    @css_class = ""
+    super
+  end
+end
```
ethn: cute.

card/mod/follow/set/type_plus_right/user/follow/follow_editor_helper.rb
```diff
@@ -105,6 +107,7 @@ def suggested_follow_option name
   # suggestion value contains only set (implies *always)
   def a_set_only_suggestion sug
     return unless (set_card = valid_set_card(sug))
+
```
ethn: general pattern is space after returns?

card/mod/follow/spec/set/all/follow/follow_link_views_spec.rb
```diff
@@ -0,0 +1,27 @@
+RSpec.describe Card::Set::All::Follow::FollowLinkViews do
+  xcontext "when not following" do
```
ethn: problems with these?

card/mod/machines/set/type/css.rb
```diff
@@ -32,6 +34,11 @@ def clean_html?
   false
 end
 
+def compress_css?
+  return true
+  !Rails.env.development?
```
ethn: config option?

card/mod/pointer/set/abstract/02_pointer/html_views.rb
```diff
@@ -5,6 +5,12 @@
     standard_pointer_core
   end
 
+  view :item_cores, cache: :never do
+    card.known_item_cards.map do |item|
+      nest item, view: :core
+    end.join "\n"
+  end
```
ethn: seems like what we want is a view that is more like `bare` and then you specify the item views.  `nest :bare, items: {view: :core}`.  I don't like the idea of different views for different item views.

card/mod/settings/set/self/create.rb
```diff
@@ -1,2 +1,3 @@
 extend Card::Setting
-setting_opts group: :permission, position: 1, rule_type_editable: false
+setting_opts group: :permission, position: 1, rule_type_editable: false,
+             help_text: "who can create cards"
```
ethn: will need localization.

ethn: let's use `# LOCALIZE` when we spot those.

card/mod/standard/set/all/rich_html/error/debug_server_error.haml
```diff
@@ -0,0 +1,1015 @@
+
+:css
+  /* Basic reset */
```
ethn: wondering if this should be in a separate mod?  kinda massive and nonstandard.

card/mod/standard/set/self/activity_toolbar_button.rb
```diff
@@ -2,13 +2,14 @@
 
 format :html do
```
ethn: delete?

card/spec/lib/card/set/format/wrapper_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        Card["A"].format_with do
+          wrapper(:cream) { "#{classy("cream")}_#{interiour}_cream" }
+
+          before(:whipped_cream_cake) { class_up "cream", "whipped" }
+          view(:whipped_cream_cake, wrap: :cream) { "cake" }
+        end
+      end
+
+      it "wraps with whipped cream" do
+        expect(format.render_whipped_cream_cake).to eq "cream whipped_cake_cream"
```
ethn: cool.

card/db/migrate/20190116170824_add_left_key_to_card_virtuals.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class AddLeftKeyToCardVirtuals < ActiveRecord::Migration[4.2]
+  def up
+    add_column :card_virtuals, :left_key, :string
+  end
+
+  def down
+    remove_column :card_virtuals, :left_key
+  end
+end
```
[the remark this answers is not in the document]
xithan: It's a standard rails migration and not a card migration. There are rollback and remigrate tasks that expect that method.

card/mod/core/layout/default.html
```diff
@@ -1,10 +1,12 @@
...
+  myhead
+</head>
+
+<body class="right-sidebar">
+  <header>{{*header|core}}</header>
+  <article>{{_main|open}}</article>
+  <aside>{{*sidebar|core}}</aside>
+  <footer>{{*footer|core}}</footer>
+</body>
 </html>
```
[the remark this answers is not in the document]
xithan: I'm not sure if html files aren't a cleaner approach than html blocks in ruby files

card/mod/standard/set/all/rich_html/error/debug_server_error.haml
xithan: A good idea, in theory, but not trivial to realize. There are related coffeescript changes in the slot stuff and changes for the ServerError exception. There is also the questions if the "shark" debugger is really useful and what to do with ajax request. I have to catch the exception if I want to show it nicely formatted in a modal but when I catch it then most of the debug features of the better_errors gem don't work. So it's either ugly unreadable error message but full better_errors features when you go afterwards to /__better_errors or nicely formatted backtrace and error message but without all the other features that /__better_errors offers like open the editor at the place where the error occurred.

card/mod/standard/set/self/activity_toolbar_button.rb
xithan: later. There is already a PT reminder for this: https://www.pivotaltracker.com/story/show/163321969

card/db/migrate/20190116170824_add_left_key_to_card_virtuals.rb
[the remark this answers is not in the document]
ethn: I meant I wasn't clear on how we're using the `card_virtuals` table in general.  (but I admit that wasn't clear!)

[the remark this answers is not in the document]
xithan: We use it to cache the content of virtual cards. We used to use left_id and right_id as identifier but that doesn't work for virtuals cards whose left part doesn't exist. *machine_cache cards are an example. Because left_id is sometimes nil for those cards and all have the same right_id some machine input cards shared the same cached content.

card/mod/edit/set/all/form.rb
xithan: parens around last three would change the logic. rubocop's auto correction removed those parens

card/mod/follow/set/type_plus_right/user/follow/follow_editor_helper.rb
xithan: yes, that's what rubocop says

PR #284, opened 2018-08-30
fix broken export views

ethn:
export_items was not returning an array of hashes. not sure since when.

### DIFF
--- a/card/mod/core/set/all/export.rb
+++ b/card/mod/core/set/all/export.rb
@@ -57,7 +57,6 @@ format :json do
   end
 
   def valid_items_for_export
-
     items_for_export.flatten.reject(&:blank?).uniq.find_all do |card|
       valid_export_card? card
     end
--- a/card/mod/core/set/all/export.rb
+++ b/card/mod/core/set/all/export.rb
@@ -8,7 +8,7 @@ format :json do
     # avoid loops
     return [] if @export_depth > max_export_depth || @exported_keys.include?(card.key)
     @exported_keys << card.key
-    Array.wrap(render_atom).concat(render_export_items).flatten
+    Array.wrap(render_export_item).concat(render_export_items).flatten
   end
 
   def max_export_depth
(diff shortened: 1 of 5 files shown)


card/mod/search/set/abstract/search/views.rb
```diff
@@ -33,26 +33,26 @@ def item_cards
...
   end
 
-  def complete_or_match_search limit=AUTOCOMPLETE_LIMIT
-    starts_with = complete_search limit
+  def complete_or_match_search limit: AUTOCOMPLETE_LIMIT
+    starts_with = complete_search limit: limit
     remaining_slots = limit - starts_with.size
     return starts_with if remaining_slots.zero?
-    starts_with + match_search(remaining_slots)
+    starts_with + match_search(not_names: starts_with, limit: remaining_slots)
```
ethn: complete_ and match_ search were returning duplicates.

card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -92,21 +92,21 @@ def inline_nests_editor?
...
-      voo.structure || voo.edit_structure || # structure configured in voo
-      card.structure ||                      # structure in card rule
-      edit_fields.present?                   # list of fields in card rule
+    voo.structure || voo.edit_structure || # structure configured in voo
+      card.structure ||                    # structure in card rule
+      edit_fields?                         # list of fields in card rule
+  end
+
+  # override and return true to optimize
+  def edit_fields?
```
ethn: (this simplification was driven by the wikirate value handling, which was the only place using `voo.editor = :nests`)

card/mod/core/set/all/export.rb
```diff
@@ -1,30 +1,54 @@
...
   def max_export_depth
     Env.params[:max_export_depth].present? ? Env.params[:max_export_depth].to_i : 2
   end
 
-  before :export_items do
-    @exported_keys = inherit(:exported_keys) || ::Set.new
+  # returns an array of Hashes (each in export_item view)
+  view :export_items, cache: :never do
+    exporting_uniques do
+      export_items_in_view :export_item
```
xithan: I'm wondering how expensive blocks are. I admit that this reads well but passing a block around only for initializing a instance variable seems like overkill. A simple method call would do the same job.

PR #286, opened 2018-09-08
new error handling

ethn:


card/mod/standard/set/all/error.rb
```diff
@@ -9,222 +9,45 @@ def copy_errors card
...
-    tr(:bad_address)
+  view :not_found, perms: :none do
+    error_name = card.name.present? ? safe_name : tr(:not_found_no_name)
+    tr(:not_found_named, cardname: error_name)
   end
 
-  view :too_deep, perms: :none, closed: true do
-    tr(:too_deep)
+  view :bad_address, perms: :none do
+    raise Card::Error::OpenError, tr(:bad_address)
```
xithan: Besides the name "OpenError" I don't like this approach conceptually. The OpenError class is kind of a contradiction in itself. An exception that is so common that we show it to the user can't be exceptional. Like this bad address here. It's not an exception. It's a very common user input. If we raise an exception then Ruby puts a lot of effort in generating debug info (like a stack trace) that we never use because there is no bug to find. The more efficient way to resurface from the deep nested block sea is to use `throw` and `catch`.

card/mod/pointer/spec/set/abstract/pointer/options_api_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.describe Card::Set::Abstract::Pointer::OptionsApi do
+  def define_options_rule type_id, content
+    Card.create! name: %i[pointer type options], type_id: type_id, content: content
```
xithan: = `create %i[pointer type options], type_id: type_id, content: content`

card/lib/card/format/error.rb
```diff
@@ -2,35 +2,34 @@ class Card
...
-        if focal?
-          focal_error e, view
-        else
-          # TODO: consider rendering dynamic error view here.
-          nested_error e, view
-        end
-      end
-
-      def debug_error e
-        raise e if Card[:debugger]&.content == "on"
```
xithan: I missed that you removed this but I don't want to miss this feature. It was the shining light that guided me through the hell of nasty bugs. Now I'm lost in the darkness of ugly formatted logger messages.

PR #296, opened 2018-10-11
Bar and box ui

srivig:


card/mod/item/set/all/bar.rb
```diff
@@ -2,6 +2,10 @@
 
 format :html do
   view :mini_bar do
+    shared = "align-items-center"
+    mini_class = "col-5 border-left d-flex justify-content-end text-align-right #{shared}"
+    class_up "bar-left", "col-7 p-2 font-weight-bold d-flex grow-2 #{shared}"
+    class_up "bar-right", mini_class
```
ethn: suppose I'm a shark that doesn't want one or more of these classes.  What do I do?

PR #303, opened 2018-11-20
hodgepodge of error handling and bugfixes.

ethn:


decko/rails/controllers/card_controller.rb
```diff
@@ -121,13 +121,20 @@ def render_page format, view
...
+  end
+
+  class << self
+    def rescue_from_class klass
+      rescue_from(klass) { |exception| handle_exception exception }
     end
   end
+
+  rescue_from_class ActiveRecord::RecordInvalid
+  rescue_from_class(Rails.env.development? ? Card::Error::UserError : StandardError)
```
xithan: What is the advantage of rescuing UserErrors?

ethn: This restores the old developer error UX.  If you don't rescue code errors, you get the full backtrace/console thing.  (But you don't really want those for user errors).

xithan: I think I got the logic wrong when I read it the first time. Makes sense now.

PR #304, opened 2018-11-24
ubos-inspired fixes

ethn:


card/db/migrate_core_cards/20160801143221_change_bootstrap_card_type_to_scss.rb
```diff
@@ -2,6 +2,7 @@
 
 class ChangeBootstrapCardTypeToScss < Card::Migration::Core
   def up
+    create_or_update name: "*machine cache", codename: "machine_cache"
```
ethn: this is needed sooner than it was appearing.  I added it in two places in case people happen to have stopped at this migration.

PR #309, opened 2018-12-07
acts refinement

ethn:
I started in on this to resolve renaming issues, specifically involving updating references after renaming.  That was resolved by handling the referer cards as subcards and making "skip" work.

But I tried to go deeper and detect and fix the error that was being raised so that other people using create/save in events didn't hit the same mysterious "Failed to save the Record" snag.  It clearly has something to do with saving a card twice in the same act, but when I built a simple test to reproduce that, it worked just fine, so I think it's a bit more complicated (eg, saving a card while another save of the same card is in progress).

The safest thing seemed to be to disallow adding the same card to the act twice, and so a lot of the changes I made were in that direction.  And I was able to avoid a good bit of duplicate act processing.  But eventually I realized that #new already added subcard directors to the act manager, so by the time save triggered the "actify" process, it would be too late to refuse to add the director because it already exists.

Anyway, I think it's cleaner, but it's not as far as I'd hoped.


card/mod/core/set/all/abort.rb
```diff
@@ -0,0 +1,75 @@
+
+# The Card#abort method is for cleanly exiting an action without continuing
```
ethn: separated all this from actify.

PR #310, opened 2018-12-17
update_attributes -> update

ethn:


card/mod/core/set/all/content.rb
```diff
@@ -104,11 +77,20 @@ def draft?
   Env.params["draft"] == "true"
 end
 
-def prepare_content
-  cont = standard_content || "" # necessary?
+def prepare_db_content
```
ethn: this is the one change that isn't related to update_attributes

card/mod/core/set/all/actify.rb
```diff
@@ -53,10 +53,10 @@ def valid?(*)
   act { super }
 end
 
-def update_attributes *args
+def update *args
   act { super }
```
xithan: Does `update_attributes` still work or is it now linked to the non-actified rails version?
Since rails supports `update_attibutes` I think we should support it, too.

ethn: In rails, `update_attributes(!)` is an alias for `update(!)`, which they changed to be the more idiomatic method.  So since we now override `update`, it will still work to call `update_attributes` –– and you'll still get an act.  This is much preferable to the confusing current situation, in which `update_attributes` gets an act, but `update` doesn't.

ethn: https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attributes

ethn:
@xithan any further concerns?

xithan:
no, only that it will produce a lot of merge conflicts with the bridge PR

card/mod/core/set/all/actify.rb
xithan: That's not true. `update_attributes` is linked to rails' `update` method not the overridden one

PR #314, opened 2019-01-17
make update_attributes work again

xithan:


card/mod/core/set/all/actify.rb
```diff
@@ -60,3 +60,6 @@ def update *args
 def update! *args
   act { super }
 end
+
+alias_method :update_attributes, :update
+alias_method :update_attributes!, :update!
```
ethn: Yes, I was wrong to say this wasn't needed. (This is how it works in rails, but we have to override their aliasing).

So now, finally, both work.

PR #317, opened 2019-02-01
bridge and cypress

xithan:
fixes a lot of bridge issues (see PT)
introduces cypress 
tests the bridge with cypress

The cypress installation comes with a lot of example specs. I moved them all to spec/cypress/examples. Although cypress is incredibly well documented I think it's helpful to have them around for a while.


card/mod/account/set/right/account.rb
```diff
@@ -176,8 +176,8 @@ def token_path_opts
     # Problem: when you do that then the fields are missing in the sign up form:
     # output( [field_nest(:email, view: :titled, title: "email"),
     #          field_nest(:password, view: :titled, title: "password")])
-    %({{+#{:email.cardname}|titled;title:email}}
-      {{+#{:password.cardname}|titled;title:password}})
+    %({{+#{:email.cardname}|edit_row;title:email}}
+      {{+#{:password.cardname}|edit_row;title:password}})
```
ethn: I was hoping to see this grow more idiomatic, but it seems to be growing less.

Is there something really special about  this that justifies all the special handling?

card/mod/account/set/right/password.rb
```diff
@@ -38,8 +38,8 @@ def ok_to_read
 end
 
 format :html do
-  view :core do
-    "<em>#{render_raw}</em"
+  view :core, wrap: :em do
```
ethn: cool.

ethn: If I were to override a view like this, could I get rid of the wrapping?

card/mod/account/set/right/account.rb
ethn: I think I'm understanding that this is the edit-in-place mechanism.  But I think we might want edit in place for a number of different views, no?  Eg, we might want something more like `{{+:email|labeled;edit:in-place}}`

(that `+:email` syntax might actually work now...)

card/mod/edit/set/all/bridge/tab_views.rb
```diff
@@ -64,9 +61,14 @@ def bridge_pills items
...
       opts = bridge_link_opts.merge("data-toggle": "pill")
-      add_class opts, "nav-link"
       opts.merge! breadcrumb_data(breadcrumb)
-      opts.merge! extra_opts if extra_opts
+      if extra_opts
+        classes = extra_opts.delete :class
+        add_class opts, classes if classes
+        opts.deep_merge! extra_opts
+      end
+      opts["data-cy"] = "#{text.to_name.key}-pill"
```
ethn: might be nice to have all this opts handling in its own method.

card/mod/account/set/right/account.rb
ethn: that said, I see how that might not be quick.  seems like we need a consistent way to denote wrapped content so we know what to edit in place.

card/mod/machines/lib/javascript/decko_slot.js.coffee
```diff
@@ -1,3 +1,60 @@
...
+#    _close-overlay
+#    _close-modal
+#
+#   To 3. Similar as 1, the slot has a few overlay and modal options.
+#     classes:
+#     _modal
+#        show slot in modal
+#     _overlay
+#        show slot in overlay
+#
```
ethn: good stuff.

decko/spec/cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+  "baseUrl": "http://localhost:5002",
+  "defaultCommandTimeout": 10000,
+  "projectId": "n4h7vq"
+}
```
ethn: Can we run this on Semaphore?

card/config/initializers/cypress_dev.rb
```diff
@@ -0,0 +1,9 @@
+if defined?(CypressDev)
```
ethn: Something I need to install?

card/mod/account/set/right/account.rb
xithan: Do you have a suggestion how we call the other two values for the edit option?  (the bridge and only the content editor in a modal) Maybe it should be a little bit more self-explanatory than "bridge".

card/mod/account/set/right/password.rb
xithan: Aren't these view tags set independent?

card/config/initializers/cypress_dev.rb
xithan: I don't think so. You only need to know the magic command that will let you enter the testing heaven.

card/mod/account/set/right/account.rb
ethn: another minor variant: `content-inline`, `content-modal`, and `full-modal`.

card/mod/account/set/right/password.rb
ethn: Looked closer. You extended the `cache` mechanism to handle these, which imo was the right way to go.  That means they're associated with the set and implemented as methods and thus overridable. So I guess all we need is to support `wrap: :none` or something if we want to be able to get rid of wrapping.  (can mark as resolved)

decko/spec/cypress.json
xithan: yes, we can and we do from now on.

ethn: nice!

PR #318, opened 2019-02-04
query, act,  and bar fixes

ethn:


card/mod/core/set/all/codename.rb
```diff
@@ -8,6 +8,8 @@ def codename
 end
 
 event :reset_codename_cache, :integrate, changed: :codename do
+  return if @action == :create && codename.nil?
```
ethn: codename cache gets cleared on every create without this

card/mod/core/set/all/name_events.rb
```diff
@@ -82,7 +82,7 @@ def changing_existing_tag_to_junction?
 end
 
 def name_in_use_as_tag?
-  Auth.as_bot { Card.count_by_wql right_id: id }.positive?
+  !Card.where(right_id: id, trash: false).take.nil?
```
ethn: Not sure this even gets used.  I don't think `changing_name_to_junction?` is ever true.

But if it _were_ used, we don't need to count them all!

card/mod/history/set/all/history/act_listing.rb
```diff
@@ -23,8 +23,8 @@ def acts_layout acts, context, draft_legend=:hide
   end
 
   def act_list acts, context
-    fmt = context == :relative ? self : act.card.format(:html)
     act_accordion acts, context do |act, seq|
+      fmt = context == :relative ? self : act.card.format(:html)
```
ethn: recent changes totally broken without this.

card/mod/search/set/abstract/wql_search.rb
```diff
@@ -103,3 +103,9 @@ def limit
     query_with_params.limit
   end
 end
+
+format :html do
+  def default_limit
+    card_content_limit || super
+  end
```
ethn: limit in WQL has no effect in HTML without this.

cardname/lib/cardname.rb
```diff
@@ -57,11 +60,6 @@ def stringify obj
       end
     end
 
-    def banned_re
-      banned_chars = (banned_array << joint).join
-      /[#{Regexp.escape banned_chars}]/
-    end
```
ethn: no need to reconstruct RE over and over again.

PR #325, opened 2019-02-14
Link2view

ethn:


card/mod/core/set/all/event_conditions.rb
```diff
@@ -8,9 +8,30 @@ def event_applies? event
...
+    if updating_type?
+      cc = Card.fetch id, skip_modules: true
+      cc.name = name
+      cc.type_id = type_id
+      cc.content = content
+      cc.include_set_modules
+    else
+      self
+    end
+end
```
xithan: 🖖 ø -> e

card/mod/core/set/all/event_conditions.rb
```diff
@@ -8,30 +8,15 @@ def event_applies? event
...
-      cc.type_id = type_id
-      cc.content = content
-      cc.include_set_modules
-    else
-      self
-    end
+# existing card is being changed in a way that alters its sets
+def updating_sets?
+  @action == :update && real? &&
+    (attribute_is_changing?("type_id") || attribute_is_changing?("name"))
```
xithan: you could use `type_id_is_changing?` and `name_is_changing` insterad

card/mod/core/spec/set/all/event_conditions_spec.rb
```diff
@@ -51,6 +51,58 @@ def change_content
...
+        with_test_events do
+          test_event :validate,
+                     on: :update, set: Card::Set::Type::SearchType, changing: :type do
+            add_to_log "NO to run"
+          end
+          update_type
+          expect(@log).to be_empty
+        end
+      end
+    end
```
xithan: How about name changes? 
I'm not sure if that's preferable but if you make the log messages more like the test description then you could put all events in one test.

card/mod/core/set/all/event_conditions.rb
ethn: ok.

PR #332, opened 2019-03-15
getting edit_in_place working

ethn:


card/lib/card/view/options.rb
```diff
@@ -46,8 +45,11 @@ module Options
           :size,           # set an image size
           :params,         # parameters for add button.  deprecated!
           :items,          # options for items (Hash)
-          :cache           # change view cache behaviour
-        ],                 #   (Symbol<:always, :standard, :never>)
+          :cache,          # change view cache behaviour
+                           #   (Symbol<:always, :standard, :never>)
+          :edit            # edit mode
```
ethn: see "NOTE" above about show/hide inheritance.

card/mod/edit/set/all/edit_in_place.rb
```diff
@@ -1,7 +1,7 @@
 format :html do
   view :edit_in_place, perms: :update, tags: :unknown_ok, cache: :never, wrap: :slot do
     with_nest_mode :edit do
-      card_form :update, success: { view: :editable } do
+      card_form :update do
```
ethn: this needs to return to the home view.

card/mod/edit/set/all/edit_in_place.rb
```diff
@@ -52,7 +48,7 @@ def edit_in_place_link view=:edit_in_place, align: :left
   end
 
   def cancel_in_place_button args={}
-    args.reverse_merge! class: "cancel-button btn-sm", href: path(view: :editable)
+    args.reverse_merge! class: "cancel-button btn-sm", href: path()
```
ethn: again, should be home view

card/mod/edit/set/all/edit_in_place.rb
```diff
@@ -16,10 +16,6 @@
      "<div class='col-sm-11'>#{render_editable}</div>"]
   end
 
-  view :editable, perms: :update, tags: :unknown_ok, wrap: :slot do
```
xithan: this view is used in the view above

card/mod/edit/set/all/edit_in_place.rb
xithan: but you have to make sure that the home view has the edit-in-place link

ethn: what? why?  If the home view didn't have a link to here, then you couldn't get here in the first place, right?

ethn: You can check out how editing works on metrics/companies on projects now to see that it's important to return to the home view.

card/mod/edit/set/all/edit_in_place.rb
ethn: Doh, missed that one. I'll just use `render_menued; edit: content_inline`, so we don't have to have separate views for different editing patterns.

card/mod/account/set/right/account.rb
```diff
@@ -174,8 +174,8 @@ def token_path_opts
     # Problem: when you do that then the fields are missing in the sign up form:
     # output( [field_nest(:email, view: :titled, title: "email"),
     #          field_nest(:password, view: :titled, title: "password")])
-    %({{+#{:email.cardname}|edit_row;title:email}}
-      {{+#{:password.cardname}|edit_row;title:password}})
+    %({{+#{:email.cardname}|labeled;title:email;edit:content_inline}}
+      {{+#{:password.cardname}|labeled;title:password;edit:content_inline}})
```
xithan: Did you check if this looks good? edit_row uses different col classes than labeled.

card/mod/edit/set/all/edit_in_place.rb
xithan: Before we introduced the edit option the editable view was the only way to get the in-place editor. I'm not questioning the change here. I'm only saying since this became much more involved (in the background) it should be tested that it works. I know that it should work but I have trust issues.

card/mod/account/set/right/account.rb
ethn: I did.  It's presentable, but as a general rule I think it would look better if only the content area (and not the whole slot) were replaced with the editor and the editor had a border that made it look more like an editor.  Will make a story...

card/mod/edit/set/all/edit_in_place.rb
ethn: Fair point. I'll write some tests.

ethn: (I did test it in the browser, fwiw, but we should definitely have some specs)

card/mod/standard/spec/set/all/rich_html/menu_spec.rb
```diff
@@ -0,0 +1,43 @@
+# -*- encoding : utf-8 -*-
+
+RSpec.describe Card::Set::All::RichHtml::Menu do
+  check_views_for_errors :edit_link, :full_page_link
```
xithan: `check_html_views_for_errors`

card/mod/standard/spec/set/all/rich_html/menu_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        expect(edit_menu)
+          .to have_tag("div.card-menu.nodblclick._show-on-hover") do
+          with_tag "a.edit-link", with: { "data-modal-class": "modal-lg",
+                                          href: "/A?view=edit" }
+        end
+      end
+    end
+
+    context "when voo.edit = :standard" do
+      specify do
```
xithan: you could save one level here
`example "voo.edit = :standard" do`

decko/lib/decko/config/environments/cucumber.rb
```diff
@@ -12,14 +12,14 @@
...
-  # config.prepopulate_cache = true
+  config.prepopulate_cache = true
 
   # Log error messages when you accidentally call methods on nil.
   config.whiny_nils = true
 
   # Show full error reports and disable caching
   config.consider_all_requests_local = true
-  config.action_controller.perform_caching             = true
+  # config.action_controller.perform_caching = true
```
xithan: we probably want that back

PR #333, opened 2019-03-19
replace minibar with bar

xithan:


card/mod/item/set/all/bar.rb
```diff
@@ -1,22 +1,25 @@
 include_set Abstract::BsBadge
 
 format :html do
-  view :mini_bar do
-    render_bar hide: [:bar_middle]
+  setting :bar_cols
+  setting :infobar_cols
+
+  view :infobar do
```
ethn: I'm probably going to trip up on the fact that expanded_bar has an underscore and infobar doesn't.  How do you feel about `info_bar`?

card/mod/item/set/all/bar.rb
```diff
@@ -1,22 +1,25 @@
...
+    class_up_bar_sides(voo.show?(:bar_middle))
     # note: above cannot be in `before`, because before blocks run before viz processing
     wrap { haml :bar }
   end
 
-  def bar_side_cols middle=true
-    middle ? [5, 4, 3] : [9, 3]
-  end
+  bar_cols 9, 3
+  infobar_cols 5, 4, 3
```
ethn: yeah, this is an improvement.

card/mod/item/set/all/bar.rb
xithan: I'm not attached to `infobar`. `bar_with_middle` would be more 
self-explanatory.

PR #337, opened 2019-03-22
tweak origin slot tracking

xithan:
I added a slot-id independent of the slot content and slot updates. That way we don't loose the connection from a modal or overlay to the origin slot after the origin slot was updated.


card/mod/standard/set/all/rich_html/html_views/info.rb
```diff
@@ -37,7 +37,7 @@
   end
 
   view :type_info do
-    return unless show_view?(:toolbar, :hide) && card.type_code != :basic
+    return unless card.type_code != :basic
```
ethn: fwiw, I think we should override this view in type/basic (but this is not new...)

card/db/migrate_core_cards/20190321085735_wagn_to_decko_logo.rb
```diff
@@ -0,0 +1,12 @@
+# -*- encoding : utf-8 -*-
+
+class WagnToDeckoLogo < Card::Migration::Core
+  def up
+    logo = Card[:logo]
+    binding.pry
```
ethn: binding.pry

PR #340, opened 2019-03-29
minor bar tweaks

ethn:
minor bar tweaks


card/mod/bootstrap/lib/stylesheets/style_bootstrap_cards.scss
```diff
@@ -380,9 +380,30 @@ body.mceContentBody {
...
+  .media {
+    width: 100%
+  }
+  .media-left {
+    max-width: 25%
+  }
+  .bar-right {
+    word-break: break-word;
+  }
+}
```
ethn: this makes bars with media more responsive

card/mod/item/set/all/bar.rb
```diff
@@ -51,23 +52,17 @@ def class_up_cols classes, cols, context=nil
 
   view :bar_left do
     class_up "card-title", "mb-0"
+    bar_title
+  end
+
+  def bar_title
```
ethn: better yet, could make this a view.

decko/spec/cypress/integration/edit/pointer.spec.coffee
```diff
@@ -56,7 +56,7 @@ describe 'editing pointers', () ->
         .should("contain", "Joe Camel")
       cy.contains(/select\s+3\s+following/)
       cy.get("input._select-all").click()
-      cy.contains(/select\s+0\s+following/)
+      # cy.contains(/select\s+0\s+following/)
```
ethn: this appears very briefly and then disappears, so it fails sporadically if the timing is wrong.

xithan: Why does it appear briefly? Shouldn't it remain 0 after I selected all?

ethn: see the `count > 0` stuff in select_item.haml for more context.

I _think_ the rationale is that if there are no results, there is no need for a checkbox.  Another way to go would be to leave the checkbox there the whole time but disable if if there are 0 items to select...

PR #342, opened 2019-04-04
layout wrapping

ethn:


card/mod/standard/spec/set/all/rich_html/overlay_spec.rb
```diff
@@ -8,8 +8,8 @@
 
     specify do
       expect(core_view)
-        .to have_tag("div.card-slot.SELF-a._overlay.d0-card-overlay") do
-        with_tag "div#main" do
+        .to have_tag("div#main") do
+        with_tag "div.card-slot.SELF-a._overlay.d0-card-overlay" do
           with_tag "span.card-title", "Z"
         end
```
ethn: I suspect there's a reason why overlay is supposed to be outside of layout. Can we articulate it?

xithan: overlay used to be a layout, even the tests says "describe overlay layout". Could be that overlays don't work anymore if there is a main div between the overlay and the overlay container

ethn: What are the cases where we use an overlay and a layout at the same time?  I see lots of cases where we use overlays in Ajax, so there is no layout (mostly in link_to_view situations).  But are these two systems ever used at the same time?  (All cukes and cyps work on both decko and wikirate with this change).

xithan: I think we use overlays only in the bridge

ethn: OK, and the bridge is well tested in cyps and cukes.

I think we're pretty safe to merge this.  It fixes some pretty clear bugs, eg on the research page. Generally speaking, I don't think we have a case yet where we really want an overlay wrapped around the main nest on a page with a layout.  Approve?

PR #346, opened 2019-04-09
final 0.7 fixes?

ethn:


card/mod/standard/set/all/rich_html/modal.rb
```diff
@@ -115,11 +115,6 @@ def close_modal_window
   end
 
   def pop_out_modal_window
-    # we probably want to pass on a lot more params than just view,
-    # but not all of them
-    # (eg we don't want layout, id, controller...)
-    popout_params = params[:view] ? { view: params[:view] } : {}
-    link_to icon_tag(:new_window), path: popout_params,
-                                   class: "pop-out-modal close"
+    link_to icon_tag(:new_window), path: {}, class: "pop-out-modal close"
```
ethn: params[:view] is almost always "bridge" in this context.  I think the most useful thing is just a clean link to the card.

card/lib/card/format/permission.rb
```diff
@@ -64,8 +64,12 @@ def task_denied_for_view view
...
       def view_for_unknown _view
         # note: overridden in HTML
-        root.error_status = 404 if focal?
-        focal? ? :not_found : :missing
+        if main?
+          root.error_status = 404
+          :not_found
+        else
+          :missing
+        end
```
ethn: this fixes cancel.

card/mod/ace_editor/lib/javascript/script_ace_config.js.coffee
```diff
@@ -32,7 +32,12 @@ $.extend decko,
     editor.setOptions conf
 
   initAce: (textarea) ->
+    return if textarea.data('aceInited')
+    textarea.data('aceInited', true)
+    # prevents double initialization
+
     mode = textarea.attr "data-ace-mode"
+
```
ethn: will get rid of this whitespace...

card/mod/standard/set/type/cardtype.rb
```diff
@@ -1,7 +1,7 @@
 include_set Abstract::WqlSearch
 
 def wql_hash
-  { type_id: id }
+  { type_id: id, sort: :name }
```
xithan: A long time ago we got rid of sorting because it was too expensive. Maybe here?

ethn: Unfortunately, card queries still sort by default:
```
        def order_config
          @mods[:sort].blank? ? "update" : @mods[:sort]
        end
```

We discussed getting rid of sorting, but I guess we never did it. One cost of losing default sorting is that our default paging behavior might get weird.  If things aren't sorting by anything in particular, how do you know the items on page 1 won't show up again on page 2?

I'm still a little iffy on the rationale for making cardtypes have their own searches rather than using associated search cards like `<Cardtype>+*type+by_name`, but I know one of the most common cases now is JSON exports, and in all the cases I was looking at for those, name sorting seemed a more useful default than date sorting.  Specifically, this made the large Company.json and Metric.json dumps easier to stay oriented in.

PR #347, opened 2019-04-13
quick fix for "add item" on filtered lists

ethn:


card/mod/pointer/set/abstract/02_pointer/html_views.rb
```diff
@@ -104,7 +104,7 @@ def add_item_modal_link
                        filter_card: filter_card.name,
                        slot_selector: filtered_list_slot_class,
                        item_selector: "_filtered-list-item",
-                       slot: { hide: :modal_footer },
+                       slot: { hide: [:modal_footer] },
```
xithan: Is this necessary? If so then it should change.

ethn: If you're asking whether hiding the modal footer necessary, I don't know the answer to that.
If you're asking whether putting the `hide` values in an array is necessary, then I think the answer is that a simple string (or symbol) will work if it's the only thing hidden, but in this case there was _another_ `hide` configuration for that slot from somewhere else.  So the params looked something like `slot[hide]=A&slot[hide][]=...`.

The problem was showing up in editing company and metric lists on wikirate project pages. This was a quick fix that got it working (it's already deployed).

card/lib/card/error.rb
```diff
@@ -58,6 +58,11 @@ def report
 
     # error whose message can be shown to any user
     class UserError < Error
+      cattr_accessor :user_error_classes
+      self.user_error_classes = [self,
+                                 ActiveRecord::RecordNotFound,
+                                 ActionController::MissingFile,
+                                 ActiveRecord::RecordInvalid]
```
ethn: This generalization fixed the problem where missing favicon files were returning 500 errors rather than 404 errors in cucumber.

card/mod/standard/set/all/path.rb
```diff
@@ -29,6 +29,10 @@
   # @option opts [Symbol] :format
   # @option opts [Hash] :card
   # @option opts [TrueClass] :no_mark
+
+  CAST_PARAMS = { slot: { hide: :array, show: :array, wrap: :array } }.freeze
```
xithan: seems at bit overengineered to me. To what else can you cast a param in a url?

card/mod/standard/set/all/path.rb
```diff
@@ -96,9 +100,31 @@ def path_extension opts
...
     opts.empty? ? "" : "?#{opts.to_param}"
   end
 
+  # normalizes certain path opts to specified data types
+  def cast_path_opts opts, cast_hash=CAST_PARAMS
+    return opts unless opts.is_a?(::Hash)
+    opts.each do |key, value|
+      next unless (cast_to = cast_hash[key])
+      opts[key] = cast_path_value value, cast_to
+    end
```
xithan: negligible worse performance (more method calls) but much nicer to read:
 `opts.transform_values!(&:cast_path_value)` and do the conditional stuff in `cast_path_value`

card/mod/standard/set/all/path.rb
ethn: I thought the casting was overkill in the first place :).

It seemed plausible to me that we might have some string representation for a hash.  And we could use something similar to force `item=bar` into `items[view]=bar`.  But meh.

card/mod/standard/set/all/path.rb
ethn: that won't work though. the cast hash depends on the key.

card/mod/standard/set/all/path.rb
ethn: My main aversion was to hard-coding slot handling in the path handling, so I was mostly trying to start down a path of configurability.

card/mod/standard/set/all/path.rb
xithan: right, ruby needs a `with_keys` method like the `with_index` method

PR #348, opened 2019-04-17
tweak double click editing

xithan:


card/mod/machines/lib/javascript/decko/autosave.js.coffee
```diff
@@ -0,0 +1,28 @@
+jQuery.fn.extend
```
xithan: no change, moved here from decko.js.coffee

card/mod/machines/lib/javascript/decko/card_menu.js.coffee
```diff
@@ -0,0 +1,15 @@
+decko.isTouchDevice = ->
```
xithan: no change, moved here from a decko.js.coffee and a separated script card  "script: card menu" that I removed

card/mod/machines/lib/javascript/decko/components.js.coffee
```diff
@@ -13,10 +15,7 @@ $(window).ready ->
       type: 'GET'
       success: (html) ->
         $(tab_id).append(html)
-        $(tab_id).triggerSlotReady()
-
-  $('body').on "select2:select", "._submit-on-select", (event) ->
-    $(event.target).closest('form').submit()
+        decko.contentLoaded($(tab_id), $(this))
```
xithan: `contentLoaded` is the new starting point for all the slot initialization stuff. I didn't name it "slotLoaded" because it is also called for modals and overlays.

card/mod/machines/lib/javascript/decko/doubleclick.js.coffee
```diff
@@ -0,0 +1,27 @@
+doubleClickActiveMap = { off: false, on: true, signed_in: decko.currentUserId }
+
+doubleClickActive = () ->
+  doubleClickActiveMap[decko.doubleClick]
+  # else alert "illegal configuration: " + decko.doubleClick
+
+doubleClickApplies = (el) ->
+  return false if ['.nodblclick', '.d0-card-header', '.card-editor', '.bridge-sidebar'].some (klass) ->
```
xithan: added bridge-sidebar

card/mod/machines/lib/javascript/decko/doubleclick.js.coffee
```diff
@@ -0,0 +1,27 @@
...
+  slot = el.slot()
+  return false if slot.find('.card-editor')[0]
+  # false if there is a card-editor open inside slot
+  slot.data 'cardId'
+
+triggerDoubleClickEditingOn = (el)->
+  slot = el.slot()
+  edit_view = decko.slotEditView(slot)
+  url = decko.path("~#{slot.data('cardId')}?view=#{edit_view}")
+  slot.reloadSlot url
```
xithan: here is the new edit mode handling for double clicks. I got rid of code duplication for ajax calls by using `reloadSlot`

card/mod/machines/lib/javascript/decko/editor.js.coffee
```diff
@@ -30,35 +35,21 @@ jQuery.fn.extend {
...
 
-doubleClickActive = () ->
-  doubleClickActiveMap[decko.doubleClick]
-  # else alert "illegal configuration: " + decko.doubleClick
+$(window).ready ->
+  decko.initializeEditors $('body')
+  # setTimeout (-> decko.initializeEditors $('body')), 10
+  # dislike the timeout, but without this forms with multiple TinyMCE editors
+  # were failing to load properly
+  # I couldn't reproduce that problem described above -pk
```
xithan: ! (first thing to check if problems with tinymce come up)

card/mod/bootstrap/lib/javascript/script_load_select2.js.coffee
```diff
@@ -19,6 +16,14 @@ $.extend decko,
         opts.minimumResultsForSearch = elem.data("minimum-results-for-search")
       elem.select2(opts)
 
+$(window).ready ->
+  $('body').on 'select2:select', '._go-to-selected', ->
+    val = $(this).val()
+    if val != ''
+      window.location = decko.path(escape(val))
```
ethn: this looks handy.

card/mod/edit/set/all/edit_inline.rb
```diff
@@ -1,11 +1,11 @@
 format :html do
-  view :edit_in_place, perms: :update, tags: :unknown_ok, cache: :never, wrap: :slot do
+  view :edit_inline, perms: :update, tags: :unknown_ok, cache: :never, wrap: :slot do
```
ethn: for another time:
1. we may want to wrap this in a frame or find another way to make sure the title is visible.
2. this should have a bridge link.

will make a story...

card/mod/machines/lib/javascript/decko/decko.js.coffee
```diff
@@ -0,0 +1,87 @@
...
+    report.show 'drop', 750
+    setTimeout (->report.hide 'drop', 750), 3000
+}
+
+#~~~~~ ( EVENTS )
+
+$(window).ready ->
+  $.ajaxSetup cache: false
+
+  $('body').on 'click', '.submitter', ->
```
ethn: would be good to prefix all these selectors with an underscore at some point...

card/mod/machines/lib/javascript/decko/doubleclick.js.coffee
ethn: why not add nodblclick to the bridge sidebar?  the other classes occur in several places, but I think the bridge-sidebar only appears once (or twice?), no?

More importantly, this may be tricky after we add guides in the bridge-sidebar. You had the good idea of making it so that edit text was double-clickable.  But the guide will be inside the bridge-sidebar.

My best guess is that we'll want to make this smarter, so that, eg, you can configure something clickable inside something not clickable, and inner configs overrule outer ones.

card/mod/machines/lib/javascript/decko/doubleclick.js.coffee
ethn: ooh, I like this.

card/mod/machines/lib/javascript/decko/editor.js.coffee
```diff
@@ -1,20 +1,25 @@
 $.extend decko,
+  initializeEditors: (range, map) ->
+    map = decko.editorInitFunctionMap unless map?
+    $.each map, (selector, fn) ->
+      $.each range.find(selector), ->
+        fn.call $(this)
```
ethn: ah.  much nicer to have this here.

card/mod/machines/lib/javascript/decko/editor.js.coffee
ethn: ✅

card/mod/machines/lib/javascript/decko/slotter.js.coffee
```diff
@@ -169,30 +168,19 @@ jQuery.fn.extend
...
       slot = slot.find(".modal-body .card-slot")
     slot.attr("data-#{type}-origin-slot-id", @closest(".card-slot").data("slot-id"))
-    #@closest(".card-slot").addClass("_#{type}-origin")
 
   updateSlot: (data, mode) ->
-    notice = @attr('notify-success')
     mode ||= "replace"
-    newslot = @setSlotContent data, mode, $(this)
-
-    if newslot.jquery # sometimes response is plaintext
```
ethn: I don't see this test anywhere.  Are we assuming we no longer do this?

xithan: `setSlotContent` checks if data is a string. In that case we don't do this.

PR #349, opened 2019-04-22
tweaks to handle decko.org performance emergency

ethn:


card/lib/card/format/nest.rb
```diff
@@ -59,13 +59,13 @@ def reuse_format?
       # recursion is caught, view recursion is not.
       # TODO: catch view recursion and remove this. (Should be straightforward within voo)
       def nest_recursion_risk?
-        content_view? && format.voo&.structure
+        content_view? # && format.voo&.structure
```
ethn: why would have to have a voo.structure to be a recursion risk??

xithan: because the structure can nest the card itself?

card/lib/card/format/nest.rb
```diff
@@ -59,13 +59,13 @@ def reuse_format?
...
       end
 
       def content_view?
         # TODO: this should be specified in view definition
         %i[
-          core content titled open closed open_content
+          bar expanded_bar core content titled open closed open_content
```
xithan: is box not a content view?

card/lib/card/format/nest.rb
ethn: If having a voo.structure indicates it's a likely risk, maybe it should be `||`.

I was seeing recursions that weren't getting caught because there was no voo.structure.

card/lib/card/format/nest.rb
ethn: good catch.

PR #353, opened 2019-04-26
view cache documentation...at long last

ethn:


card/mod/basic_formats/set/all/json.rb
```diff
@@ -27,7 +27,7 @@ def show view, args
     JSON.send method, raw
   end
 
-  view :status, unknown: true, perms: :none, cache: :never do
+  view :status, unknown: true, perms: :none do
```
ethn: I know you're not going to like my taking `cache: :never` away, but I want to understand and explain every case where we have it turned off.

PR #354, opened 2019-05-01
Cache documentation and code organization

ethn:
This looks like more change than it is.  I was mostly organizing `Card::View` modules into options, cache, and other.


card/lib/card/view/cache.rb
```diff
@@ -31,58 +31,73 @@ class View
...
     #
-    #        # unsafe, because x is manipulated
+    #        # unsafe and thus never cached, because x is altered
     #        view(:z, cache: :never) { render_z.reverse }
     #
-    # 2. view performs a dynamic search (eg via Card.search) whose results may change
-    #
-    # 3. view displays a timestamp
+    # Specifically, the danger is that the inner view will be rendered as a stub,
+    # and the out view will end up altering the stub and not the view.
```
xithan: ok, that explains it. Kind of. ~~Why does `render_x` return a stub? I thought we stub unsafe stuff like nests. Isn't another view of the same card safe and we get the cached view?~~

A thought: If `render_x` can return a stub is it maybe better to have api to "force render" (never give me a stub) instead of marking the parent view as unsafe?

card/lib/card/view/cache.rb
```diff
@@ -31,58 +31,73 @@ class View
...
     #
-    # 4. view manipulates instance variables or environmental variables that may affect
-    #    nested views and/or cards
+    # Although alterations should be considered dangerous, they are actually only
+    # problematic in situations where the inner view might sometimes render a stub.
+    # If the outer view is rendering a view of the _same card_ with all the _same view
+    # settings_ (perms, unknown, etc), there will be no stub and thus no error.
+    # Remember, however, that a view on a narrow set may inherit view settings
+    # from a general set. To be confident that a view alteration is safe, all inherited
+    # settings must be taken into account.
```
xithan: I should read all the documentation before start making comments.

card/lib/card/view/cache.rb
ethn: if there's some reason to render x as a stub, that means we think it's not safe to cache it in another view. So a forcing a render and then caching it is more dangerous than not caching.

card/lib/card/view/cache.rb
```diff
@@ -31,58 +31,73 @@ class View
...
     #        # obviously safe to cache
     #        view(:x) { "ABC" }
     #
-    #        # also safe, because x is not manipulated
+    #        # also safe, because x is NOT altered
     #        view(:y) { render_x + "DEF" }
     #
-    #        # unsafe, because x is manipulated
+    #        # unsafe and thus never cached, because x is altered
     #        view(:z, cache: :never) { render_z.reverse }
```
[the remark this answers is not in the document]
ethn: good catch.

PR #355, opened 2019-05-03
towards full yard documentability

ethn:


card/lib/card/mod/module_template.rb
```diff
@@ -33,33 +41,45 @@ def simple_load?
...
+        content_lines.each do |line|
+          comment?(line) ? comment_lines << content_lines.shift : break
+        end
+
+        @content = content_lines.join "\n"
+        @module_comment = comment_lines.join "\n"
+      end
+
+      def comment? line
+        line.match?(/^ *\#/)
```
xithan: Have you considered `#! no set module` comments in the first line?  (idk if they can appear in this place or if it's a problem to move them)

xithan: Or utf-8 comments?

ethn: I noted this in slack, but I'm confused about `no set module`. It kinda looks like it isn't actually doing anything right now.

As to your question: I'm not entirely sure, but I know:
1. none of the comment stuff takes any effect unless you're doing tmp files. 
2. usually we add the utf8 stuff automatically after all the processing.
3. my _guess_ is that yardoc ignores `#!` 

So, it's possible that something non-ideal happens in tmp files, but maybe not, and in any case it's an edge case.  My inclination is not to worry about it until we see something we don't like in a doc.

ethn:
Re the i8n changes, basically it was breaking when I loaded with tmp files. It was incorrectly inferring the mod scope from the tmp file.  There was some code for that, but it didn't work.

PR #357, opened 2019-05-06
reversion fixes.

ethn:


card/mod/history/lib/card/act.rb
```diff
@@ -73,10 +73,13 @@ def actor
     # the act's primary card
     # @return [Card]
     def card
-      res = Card.fetch card_id, look_in_trash: true, skip_modules: true
-      return res unless res&.type_id&.in?([FileID, ImageID])
+      Card.fetch card_id, look_in_trash: true #, skip_modules: true
 
-      res.include_set_modules
+      # FIXME: if the following is necessary, we need to document why.
+      # generally it's a very bad idea to have type-specific code here.
```
xithan: Why is it bad? What's the alternative? This is an AR class not a card set and `res` doesn't include set modules.

If you check the git logs you see that this was introduced to fix broken files and images in the history. Since `skip_modules: true` was probably there only for performance reasons it should be safe to remove it.

ethn: it's bad because as soon as anybody creates a new type/Image class, it will break.  Generally speaking, if we hardcode type handling, we handcuff monkeys.

ethn: Agree re performance.  I suspect (but have not confirmed) that this was the ultimate source of the reversion error.  (Though the error itself was not as big a deal as the over-reversion.

card/mod/history/lib/card/act/act_renderer/relative_act_renderer.rb
```diff
@@ -46,6 +46,11 @@ def current_act?
         def actions
           @actions ||= @act.actions_affecting(@card)
         end
+
+        def revert_link
+          revert_link_to "this", revert_actions: actions.map(&:id),
```
ethn: because we call `actions` here (as opposed to `act.actions` in the format), this is now limited only to the relative actions desired.

card/mod/standard/set/all/rich_html/title.rb
```diff
@@ -1,13 +1,27 @@
...
+    name_variant title_in_context(voo.title)
+  end
+end
+
 format :html do
-  view :title do
-    title = wrapped_title super()
-    title = link_to_card card.name, title if show_view? :title_link, :hide
+  # NOCACHE because alters @context_names
+  view :title, cache: :never do
```
ethn: updated this to adhere to our new caching best practices.

ethn: it would probably be smarter to make the name context handling part of the view api eventually.

PR #358, opened 2019-05-08
paging, item, and name context fixes

ethn:


card/lib/card/env.rb
```diff
@@ -45,19 +45,12 @@ def slot_opts
         # FIXME:  upgrade to safe parameters
         self[:slot_opts] ||= begin
           opts = params[:slot] || {}
-          opts.merge shortcut_slot_opts
+          opts.merge! shortcut_slot_opts
```
ethn: important bang.

ethn: with this fixed, we can get rid of all the other references to `params[:item]`

card/mod/standard/set/all/rich_html/show.rb
```diff
@@ -11,7 +11,7 @@ def show_method
   def show_without_page_layout view, args
     @main = true if params[:is_main] || args[:main]
     args.delete(:layout)
-    view ||= args[:home_view] || :open
+    view ||= args[:home_view] || default_nest_view
```
ethn: this is debatable.

card/mod/standard/set/all/rich_html/process_layout.rb
```diff
@@ -28,6 +28,7 @@ def page_layout
   def render_with_layout view, layout, args={}
     view_opts = Layout.main_nest_opts(layout, self)
     view ||= view_opts.delete(:view) || default_nest_view
+    view_opts[:home_view] = view
```
ethn: needed for paging.

card/lib/card/env.rb
xithan: If you want to manipulate params maybe you should do it less concealed.

ethn: I don't think we need to manipulate params.  I'll change this to clone.

PR #359, opened 2019-05-09
 move permission handling to voo

ethn:


card/lib/card/view/permission.rb
```diff
@@ -0,0 +1,58 @@
...
+    #
+    # Views can be configured in {Set::Format::AbstractFormat#view view definitions}
+    # with the `perms` directive, eg
+    #
+    #         # only render if user has permission to update card
+    #         view :myview, perms: :update do...
+    module Permission
+      private
+
+      def approve_view
```
ethn: if you open the diff to the deleted format/permission.rb, you'll see a satisfying reduction of `view` arguments.

card/mod/standard/set/all/rich_html/modal.rb
```diff
@@ -72,7 +72,6 @@ def modal_submit_button opts={}
     modal_link _render_title, size: voo.size
   end
 
-  # @param size [:small, :medium, :large, :full] size of the modal dialog
```
ethn: maybe we need a comment of this sort?  But there is no `size` param here, and yardoc was complaining.  And I don't know what the right comment would be...

card/mod/core/set/all/states.rb
```diff
@@ -1,13 +1,12 @@
 module ClassMethods
-  def exists? mark
-    card = quick_fetch mark
-    card.present?
+  def real? mark
```
ethn: I think we should deprecate `exists?` and only use `real?`.  More consistent.

card/lib/card/view/permission.rb
```diff
@@ -0,0 +1,54 @@
...
+        view ||= requested_view
+        format.view_setting setting_name, view
+      end
+
+      # by default views can't handle unknown cards, but this can be overridden in
+      # view definitions with the `unknown` directive
+      def alter_unknown
+        setting = setting(:unknown)
+        return if setting == true || card.known?
+        setting.is_a?(Symbol) ? setting : format.view_for_unknown(requested_view)
```
ethn: this means you can now do, eg, `unknown: :blank`.  Will document...

card/lib/card/format/error.rb
```diff
@@ -1,6 +1,40 @@
 class Card
   class Format
     module Error
+      def ok? task
```
xithan: We don't have format based permission rules. Why make it a format method? 

In my pending PR I have to check permissions for another card. With this change I have to create a format object for that (https://github.com/decko-commons/decko/pull/360/files#diff-2ed9b8393ffd5a7e60ce5dc1ec38ddeeR63)

ethn: This is actually not a change.  It's just moved here from format/permission.rb (the rest of which was moved into card/view/permission.rb).

The real permission handling is still on cards.  This method's main job is caching.  This way you don't have to look up the same permission many times for the same format.  

The create/update logic is a little nuanced here (again, not new, but we can revisit it and should probably comment on it).  It's saying that if a view requires update permission but the card is new, then you should have to have create permission to see it.  

In principle we could do this local caching on the card instead of the format, but we'd need to make sure we cached the as_user, too.  (I don't think we ever use "as_bot" in view rendering; we use the `_render` (or `skip_perms`) pattern instead, so we don't have to have it here).

card/lib/card/format/error.rb
```diff
@@ -1,6 +1,40 @@
...
+        task = :create if task == :update && card.new_card?
+        @ok ||= {}
+        @ok[task] = card.ok? task if @ok[task].nil?
+        @ok[task]
+      end
+
+      def anyone_can? task
+        return false unless task.is_a? Symbol
```
ethn: also just for caching to optimize multiple views in a single format. See `active_cache_permissible?`

card/mod/core/set/all/states.rb
xithan: but more inconsistent with Rails

ethn: ok, maybe we don't deprecate, but we do make `real?` our best practice? 

Alternatively, we could go the other way and deprecate `:real` as a status in favor of `:exists`

PR #361, opened 2019-05-16
filter and sort

ethn:


card/db/migrate_core_cards/20190424100531_update_credit_image.rb
```diff
@@ -66,7 +66,7 @@ def svg attr=""
   end
 
   def up
-    credit = Card[:credit_image]
+    credit = ensure_card "*credit_image", codename: :credit_image
```
ethn: wikirate didn't have this.

card/db/migrate_core_cards/20190424101004_add_read_anyone_rules.rb
```diff
@@ -2,11 +2,12 @@
 
 class AddReadAnyoneRules < Card::Migration::Core
   def up
-    cards = [:title, "*main menu", :navbox, :credit, :credit_image].map { |c| [c, :self] }
-    cards << [:head, :right]
+    ensure_card "*main menu", codename: :main_menu
+    ensure_card "*credit", codename: :credit
```
ethn: ...or this.

card/mod/machines/lib/javascript/decko/filter.js.coffee
```diff
@@ -1,180 +1,99 @@
+#decko.slotReady (slot) ->
+#  slot.find("._filter-widget").each ->
+#    showDefaultFilters $(this)
+
 $.extend decko,
```
ethn: two main changes in this file:
1. moved all the filtered list (filter items) stuff to a separate file.
2. restructured the filter handling to support resets.

The reset change involves starting with a full list of clean prototypes. Before there were only prototypes of hidden filters.  Now there's a full list, and we clone and initialize and show the active ones when the page is loaded.

card/mod/machines/lib/javascript/decko/filter_items.js.coffee
```diff
@@ -0,0 +1,125 @@
+# FILTERED LIST / ITEMS INTERFACE
+# (fancy pointer ui)
```
ethn: didn't intentionally change anything in here.

card/mod/machines/set/self/script_decko.rb
```diff
@@ -1,11 +1,11 @@
 include_set Abstract::CodeFile
 
 def source_files
-  %w[mod editor name_editor autosave doubleclick layout navbox upload filter
+  %w[mod editor name_editor autosave doubleclick layout navbox upload
      slot modal overlay recaptcha
      slotter bridge
      nest_editor nest_editor_rules nest_editor_options nest_editor_name
-     components decko follow card_menu slot_ready].map do |n|
+     components decko follow card_menu slot_ready filter filter_items].map do |n|
```
ethn: needed slotReady.  (surprised nothing else up there needs it)

card/mod/search/set/abstract/00_filter_helper.rb
```diff
@@ -29,11 +29,9 @@ def safe_sql_param key
 end
 
 def filter_keys_with_values
-  (filter_keys + advanced_filter_keys).map do |key|
+  filter_keys.map do |key|
```
ethn: unless I misunderstand, `advanced_filter_keys` doesn't do anything any more; all the magic is now in `default_filter_option`

decko/spec/cypress/integration/edit/pointer.spec.coffee
```diff
@@ -47,7 +47,7 @@ describe 'editing pointers', () ->
     cy.visit("/Joe User+friends")
     cy.get("._add-item-link").click()
     cy.contains("Select Item")
-    cy.contains("button", "Add filter").click()
+    cy.contains("button", "More Filters").click()
```
ethn: debatable. I think "more filters" is less scary, because users don't think they're being asked to add anything.  But "add filter" makes more sense if you have an initial setup with no filters on.

card/mod/machines/lib/javascript/decko/filter.js.coffee
```diff
@@ -1,187 +1,122 @@
...
-    addFilterDropdown = $selected_item.closest("._add-filter-dropdown")
-    category = $selected_item.data("category")
-    widget = addFilterDropdown.closest("._filter-widget")
-    removeCategoryOption(addFilterDropdown, category)
-    showFilterInputField(category, widget)
+#decko.slotReady (slot) ->
+#  slot.find("._filter-widget").each ->
+#    showDefaultFilters $(this)
+
+decko.slotReady (slot) ->
```
ethn: Two main changes in this file:
1. moved all the filtered list (filter items) stuff to a separate file.
2. restructured the filter handling into more object oriented structure to support resets and other external filter manipulations.

The reset change involves starting with a full list of clean prototypes. Before there were only prototypes of hidden filters.  Now there's a full list, and we clone and initialize and show the active ones when the page is loaded.

card/mod/machines/lib/javascript/decko/filter.js.coffee
```diff
@@ -1,187 +1,155 @@
-$.extend decko,
-  filterCategorySelected: ($selected_item) ->
-    addFilterDropdown = $selected_item.closest("._add-filter-dropdown")
-    category = $selected_item.data("category")
-    widget = addFilterDropdown.closest("._filter-widget")
-    removeCategoryOption(addFilterDropdown, category)
-    showFilterInputField(category, widget)
+decko.slotReady (slot) ->
+  slot.find("._filter-widget").each ->
+    if slot[0] == $(this).slot()[0]
```
ethn: THIS is the reason why I want some API like "slot loaded".

If `div._filter-widget` is in a slot within a slot, then I have to do something like this to prevent the code from being called twice.

I suspect there are already many cases where we are getting repeat calls for this.  In this case it breaks if there are repeat calls, but even if it doesn't break, it's wasteful.

xithan: It's easy to get the closest slot if needed. If you check all slotReady calls I'm fine with changing it so that it is only triggered for the root slot and not for nested slots.

ethn: When you say "it's easy to get the closest slot", you're talking about the outermost slot(s) in any given call, right?  (As opposed to "closest" in the jquery sense, meaning there's some better way to do what I've done here)

PR #363, opened 2019-05-20
fix "pristine" migration

xithan:


card/mod/core/set/all/states.rb
```diff
@@ -76,7 +76,9 @@ def new?
 
 # has not been edited directly by human users.  bleep blorp.
 def pristine?
-  new_card? || !actions.joins(:act).where(
-    "card_acts.actor_id != ?", Card::WagnBotID
-  ).exists?
+  new_card? || !user_changes?
```
ethn: so much nicer to read.

card/lib/card/migration.rb
```diff
@@ -102,6 +102,13 @@ def merge_cards names_or_keys
       Card::Migration::Import.new(data_path).merge only: names_or_keys
     end
 
+    def merge_pristine_cards names_or_keys
+      names_or_keys = Array(names_or_keys)
+
+      pristine = names_or_keys.select { |n| !Card.exists?(n) || Card.fetch(n)&.pristine? }
```
ethn: isn't the `!Card.exists?` unnecessary  here?

ethn: Oh, I guess you'd need to fetch with "new", eg:
`Card.fetch(n, new: {}).pristine?`

xithan: Yes, I noticed that that would work while I was fixing this but it didn't feel like an improvement.
(the real issue was the missing ! on the select call)

PR #370, opened 2019-06-06
media snippet cleanup.

ethn:


card/lib/card/cache.rb
```diff
@@ -26,6 +26,8 @@ class Cache
     cattr_reader :cache_by_class
 
     class << self
+      attr_accessor :no_renewal
```
ethn: see notifications spec

card/mod/follow/spec/set/all/notify_spec.rb
```diff
@@ -136,6 +136,17 @@ def notification_email_for card_name, followed_set: "#{card_name}+*self"
...
 
   describe "#notify_followers" do
+    # Normally, delayed events renew the cache, which entails clearing the local cache.
+    # That breaks these tests, because if the local cache is cleared, then the +*account
+    # card that receives the :send_change_notice method is a _different_ object than
+    # the one expecting it.
+    #
+    # An alternative approach would be to insert _only_ the account object into the
+    # cache when the delayed job is started.  This would more reliably test that the
+    # delayed job can get everything it needs with a clean cache.
```
ethn: This may be why we changed the renew call in the first place.

I'm even more confident now that #renew is right, because I can see how it would break this spec but not because it actually broke the functionality.  (In fact I tested and it was working correctly, but the spec couldn't tell).

card/mod/carrierwave/set/type/image/html_views.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+  end
+
+  def new_image action
+    card.with_selected_action_id action.id do
+      Card::Content::Diff.render_added_chunk _render_core
+    end
+  end
+
+  view :boxed, unknown: true do
```
ethn: most of this file is just moved, but the boxed and boxed link views are new.

card/mod/machines/lib/javascript/decko/decko.js.coffee
```diff
@@ -64,10 +64,14 @@ $(window).ready ->
...
-  $('card-view-placeholder').each ->
-    $this = $(this)
-    $.get $this.data("url"), (data, status) ->
-      $this.replaceWith data
+decko.slotReady (slot) ->
+  slot.find('card-view-placeholder').each ->
+    $place = $(this)
+    return if $place.data("loading")
+
+    $place.data "loading", true
```
ethn: another example of needing special handling to avoid multiple slotReady calls.

card/mod/utility/set/abstract/media.rb
```diff
@@ -21,13 +12,6 @@ def text_with_image opts={}
...
   def normalized_text_with_image_opts opts
-    opts.reverse_merge! title: _render_title,
-                        text: "",
-                        src: image_src(opts),
-                        alt: image_alt,
-                        size: :original,
-                        media_opts: {},
-                        media_left_extras: "",
-                        media_left: ""
+    opts.reverse_merge! title: _render_title, text: "", size: voo.size, media_opts: {}
```
xithan: No more `src` parameter for image tags doesn't seem right to me.

ethn: See media_snippet.haml below. Now we're nesting the image card and configuring as necessary with `items`.  The image-box is part of the image view. 

Note that this approach has allowed a lot more standardization (eg with thumbnails), less code overall, and it made it really simple to have the new "default" images (actually icons) on wikirate.

ethn: Also, supporting "src" is begging for caching issues.  nesting is the decko way.

xithan: Does it still look good if there is no image? I think that it was hard work for Vignesh to achieve that.

card/mod/pointer/set/abstract/00_paging_params.rb
```diff
@@ -1,3 +1,5 @@
+MAX_ANONYMOUS_SEARCH_PARAM = 1000
```
ethn: This was an emergency patch that I put out last night after seeing that there were a lot of requests with extremely high offsets that were taking a super long time to complete.  Not a durable solution.

card/lib/card/env.rb
```diff
@@ -47,7 +47,7 @@ def slot_opts
           opts = params[:slot]&.clone || {}
           opts = opts.to_unsafe_h if opts.is_a? ActionController::Parameters
           opts.merge! shortcut_slot_opts
-          opts.deep_symbolize_keys
+          opts.deep_symbolize_keys.slice(*Card::View::Options.shark_keys)
```
ethn: This was another emergency patch. It prevents `skip_perms` from being used in params.  It's possible it prevents something we actually want, too.

card/mod/utility/set/abstract/media.rb
ethn: yes.  I'm not sure he had mastered the unknown handling (the api for which has since improved).

PR #372, opened 2019-06-17
default filter hash handling

ethn:


card/mod/search/set/abstract/00_filter_helper.rb
```diff
@@ -13,7 +13,7 @@ def filter_param field
 # #filter_hash is the _value_ of the hash with :filter as the key.
 def filter_hash
   @filter_hash ||= begin
-    filter = Env.params[:filter]
+    filter = (Env.params[:filter] || default_filter_hash).clone
     filter = filter.to_unsafe_h if filter&.respond_to?(:to_unsafe_h)
```
xithan: This seems to be a little off in the case that we are dealing with an ActionController::Parameters object. 
I don't know if and how `clone` works in that case but I don't think we need it because `to_unsafe_h` give us already a new independent object.

ethn: good point.

PR #374, opened 2019-06-25
edit permissions can be specific to Role

xithan:
`"Role"+*members+*self+*update` rules define who can assign "Role" roles to other users


card/lib/card/format/render.rb
```diff
@@ -122,12 +122,14 @@ def view_setting setting_name, view
...
-          end
+        stub_debugging do
+          expand_stubs cached_content
         end
-        if result =~ /stub/
+      end
+
+      def stub_debugging
+        result = yield
+        if Rails.env.development? && result =~ /stub/
```
ethn: if we keep this, we should add a config option rather than always debugging in development.

card/mod/navbar/set/self/account_links.rb
```diff
@@ -8,15 +8,20 @@ def ok_to_read
   view :core, cache: :never do
     status_class = Auth.signed_in? ? "logged-in" : "logged-out"
     wrap_with :span, id: "logging", class: status_class do
-      item_links.join " "
+      navbar_items.join "\n"
```
ethn: starting to think we want an "item_joint" view option.  Maybe "separator" (to use the language of csvs?)

card/mod/rules/set/rule/table_row.rb
```diff
@@ -18,7 +18,7 @@
   end
 
   def wrap_closed_rule rule_card
-    klass = rule_card&.real? ? "known-rule" : "missing-rule"
+    klass = rule_card&.real? ? "known-rule" : "table-active"
```
ethn: what does this do?

xithan: use bootstrap styling for rule tables

PR #376, opened 2019-07-03
Fix role interface

xithan:


card/lib/card/env.rb
```diff
@@ -55,6 +55,14 @@ def session
         self[:session] ||= {}
       end
 
+      def reset_session
+        if session.is_a? Hash
+          self[:session] = {}
+        else
+          self[:controller]&.reset_session
```
ethn: What actually changed to make the session resets start working?

ethn: oh...were the "current roles" the real problem?

xithan: yes, @current_roles survives across requests but only in production mode

PR #378, opened 2019-07-10
pointer overhaul and homepage tweaks

xithan:


card/db/migrate_core_cards/20190710090205_homepage_tweaks.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+
+class HomepageTweaks < Card::Migration::Core
+  def up
+    merge_cards ["*getting started", "*main menu", "*getting started+links"]
+    update_card :wagn_bot, name: "Decko Bot"
+    delete_code_card :customized_skin
+    delete_code_card :bootswatch_theme
```
xithan: Those are both cardtypes. I'm not sure if there are no cards of that type out in the world but we don't use them anymore.

PR #381, opened 2019-07-18
tweaks to subcard handling

ethn:


card/lib/card/act_manager.rb
```diff
@@ -107,9 +107,19 @@ def clear
         directors.each_pair do |card, _dir|
           card.director = nil
         end
+        expire
         @directors = nil
       end
 
+      def expire
+        expirees.each { |expiree| Card.expire expiree }
+        @expirees = []
```
ethn: only expire at the very end of the act.

could also happen at the very end of storage, so long as it's after all the subcards have finished  the storage phase, too.

card/spec/lib/card/subcards_spec.rb
```diff
@@ -153,5 +153,18 @@ def local_content name
...
+    it "handles errors on rename" do
+      with_test_events do
+        test_event :finalize, on: :update do
+          if name == "X" # X refers to T
+            errors.add :content, "referrer error!"
+          end
+        end
+
+        expect { Card["T"].update! name: "Tea", update_referers: true }
+          .to raise_error(Card::Error::ServerError)
```
ethn: This was not only not raising an error but successfully completing the save action (and proceeding to integration)

card/mod/core/set/all/update_read_rules.rb
```diff
@@ -6,25 +6,10 @@
...
-      Card.expire name
-    end
+    structuree_names.each { |name| ActManager.expirees << name }
   end
 end
-
-event :expire_related_names, before: :expire_related, changed: :name do
-  # FIXME: look for opportunities to avoid instantiating the following
-  descendants.each(&:expire)
-  name_referers.each(&:expire)
```
ethn: Tests may prove me wrong, but I think this should be unnecessary now that descendants and name_referers are both handled in subcards (which will expire themselves).

card/mod/core/set/all/name_events.rb
```diff
@@ -14,6 +14,7 @@
 event :validate_name, :validate, on: :save, changed: :name do
   validate_legality_of_name
   return if errors.any?
+  Card.write_to_soft_cache self
```
ethn: Make sure we have the current card in the cache (this wasn't happening in Card.create).

I do it here, because I think it only really helps if we do it when we have a valid name.  But there's probably a smarter way...

PR #383, opened 2019-07-25
introduce guide rules

xithan:


card/mod/standard/set/all/rich_html/html_views/guide.rb
```diff
@@ -0,0 +1,51 @@
...
+    return "" unless guide_text.present?
+
+    alert(:secondary, true, false, class: "guide") { guide_text }
+  end
+
+  def raw_guide_text
+    false
+  end
+
+  def rule_based_guide
```
ethn: seems like this would be easier to read like this:
```
with_nest_mode :normal do
  if raw_guide_text
    process_content raw_guide_text, chunk_list: :references
  elsif card.guide_card
    nest card.guide_card, view: :core
  else
    "
end

PR #385, opened 2019-08-20
Improve cardtype+*type+*structure

xithan:


card/lib/card/content.rb
```diff
@@ -18,8 +18,7 @@ class Content < SimpleDelegator
     attr_reader :revision, :format, :chunks, :opts
 
     # initialization parses String, detects chunks
-    # @param content [String]
-    # @param format_or_card [Card::Format or Card]
+    # @param content [Strins [Card::Format or Card]
```
ethn: was this intentional?

PR #394, opened 2019-09-02
move image-box styling from wikirate to decko (card)

ethn:


card/db/migrate_core_cards/20190823220018_cleanup_for_decko_1_0.rb
```diff
@@ -9,7 +9,7 @@ def up
       delete_card [perm, :right, :options]
     end
     delete_card "*missing link"
-    delete_card "tags"
+    # delete_card "tags"
```
ethn: WikiRate has 8000+ `+tag` cards.  We don't want tag in the seed data, but I'm not sure we want to migrate it out of every site.

card/mod/bootstrap/set/abstract/bootswatch_theme/thumbnail.haml
```diff
@@ -1,10 +0,0 @@
-.card.mb-4.box-shadow
```
ethn: there is no thumbnail_image view in decko now.  I think this haml file must have been obviated?

card/mod/media/lib/stylesheets/style_image_box.scss
```diff
@@ -0,0 +1,79 @@
+//images
```
ethn: from wikirate

card/db/migrate_core_cards/20190823220018_cleanup_for_decko_1_0.rb
xithan: right, we don't want that

PR #395, opened 2019-09-03
Input type upgrade

ethn:


card/lib/card/view/options.rb
```diff
@@ -51,7 +51,7 @@ module Options
           :title,          # overrides the name of the card
           :variant,        # override the canonical version of the name with a different
                            # variant
-          :editor,         # inline_nests makes a form within standard content (Symbol)
+          :input_type,     # inline_nests makes a form within standard content (Symbol)
```
xithan: In theory this needs a migration

card/mod/pointer/set/abstract/02_pointer/html_views/radio_input.haml
```diff
@@ -11,4 +11,3 @@
         = radio_button_tag input_name, option_name, checked,
                            id: id, class: "pointer-radio-button #{'_submit-on-change' if submit_on_change}"
         = option_label option_label, id
-        = option_description "radio", option_name
```
xithan: I don't like it that we make it impossible additional help. At least we should offer a view that gives access to additional help. For example with a question mark tooltip.

card/lib/card/view/options.rb
ethn: Yes, but in practice I'm not sure anybody but us knows about it, and in WikiRate we only use it in code (I think).

card/mod/pointer/set/abstract/02_pointer/html_views/radio_input.haml
ethn: I agree.  This gets quickly into sharking views...

PR #397, opened 2019-09-06
Tweak nest editor

xithan:


card/mod/rules/set/type/set.rb
```diff
@@ -80,3 +80,14 @@ def related_sets with_self=false
     left(new: {}).related_sets with_self
   end
 end
+
+def left_type_for_nest_editor_set_selection
+  return unless  Card.fetch_id(card.name.tag).in?([StructureID, DefaultID])
+
+  set_pattern_id = Card.fetch_id card.name.left.tag
```
ethn: This is the type/set set.  The set_pattern is the tag, right?  This looks like it's written for a rule card.

card/mod/rules/set/type/set.rb
```diff
@@ -80,3 +80,14 @@ def related_sets with_self=false
     left(new: {}).related_sets with_self
   end
 end
+
+def left_type_for_nest_editor_set_selection
+  return unless  Card.fetch_id(card.name.tag).in?([StructureID, DefaultID])
```
ethn: how could the tag of a set be either of those?

card/mod/rules/set/type/set.rb
ethn: in any case, we should probably be using methods on the card like `def set_pattern_id` or `def set_pattern_code` so that we don't have to think so hard.

decko/spec/cypress/cypress_helper.rb
```diff
@@ -6,25 +6,28 @@
...
 
-# CypressDev::SmartFactoryWrapper.configure(
+# CypressOnRails::SmartFactoryWrapper.configure(
 #     always_reload: !Rails.configuration.cache_classes,
 #     factory: factory,
-#     files: %w(spec/factories.rb ./spec/factories/**/*.rb)
+#     files: [
+#       Rails.root.join('spec', 'factories.rb'),
+#       Rails.root.join('spec', 'factories', '**', '*.rb')
+#     ]
```
ethn: really?  was this just supposed to be temporary?

card/mod/rules/set/type/set.rb
ethn: I addressed these things in my PR

PR #405, opened 2019-09-12
fix view caching bug in help links

ethn:


card/mod/core/set/all/codename.rb
```diff
@@ -16,7 +16,8 @@ def codename
 private
 
 def validate_codename_permission
-  return if Auth.always_ok?
+  return if Auth.always_ok? || Auth.as_id == creator_id
```
ethn: this is so that Urs can add codenames for his metrics. Not a super well-considered change, but probably not one that will cause many problems.

decko/lib/decko/application.rb
```diff
@@ -21,7 +21,7 @@ class Application < Rails::Application
     end
 
     initializer :load_card, after: :load_config_initializers, group: :all do
-       Card
+      Card if Cardio.load_card?
```
ethn: This was causing problems (newly exposed by the decko command line exit handling) in calls processed before the database is created.

PR #406, opened 2019-09-18
view stub improvements

ethn:
use msgpack (faster and more compact than JSON)
get rid of extra escaping

### DIFF
--- a/card/lib/card/view/cache/stub.rb
+++ b/card/lib/card/view/cache/stub.rb
@@ -1,3 +1,5 @@
+require "msgpack"
+
 class Card
   class View
     module Cache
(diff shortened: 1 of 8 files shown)


card/lib/card/view/cache/stub.rb
```diff
@@ -9,30 +11,11 @@ module Cache
...
-            stub_json.gsub "(", "_OParEN_"
-          end
-
-          def unescape stub_json
-            stub_json.gsub "_OParEN_", "("
-          end
-
-          # FIXME: escaping and unescaping stubs should not be necessary
-          # It's used to avoid problems with altered views, but altering views is
-          # unsafe and should be eliminated.  See {Card::View::Cache}.
```
ethn: I don't think this comment is correct. I think the reason we had to escape/unescape parens was that we were matching on `[^)]` in view_stub.rb

card/mod/core/chunk/view_stub.rb
```diff
@@ -1,11 +1,13 @@
...
 class Card
   class Content
     module Chunk
       class ViewStub < Abstract
         Chunk.register_class(
           self,
-          prefix_re: Regexp.escape("(stub)"),
-          full_re: /\A\(stub\)([^\(]*)\(\/stub\)/,
+          prefix_re: Regexp.escape("(StUb"),
+          full_re: /\A\(StUb(.*?)sTuB\)/,
```
ethn: `?` = nongreedy here.

PR #409, opened 2019-09-20
conditional field settings

xithan:


xithan:
we do: https://github.com/decko-commons/decko/pull/409/files#diff-91e1b7df809483915d726cab13fe29fa

PR #410, opened 2019-09-26
view stub regexp and other delights

ethn:


card/mod/bootstrap/lib/javascript/script_load_select2.js.coffee
```diff
@@ -6,20 +6,27 @@ decko.slotReady (slot) ->
...
+    $(this).deInitSelect2()
 
 $.extend decko,
   initSelect2: (elem) ->
     if elem.length > 1
       decko.initSelect2($(single_el)) for single_el in elem
     else
       opts = { dropdownAutoWidth: "true", containerCssClass: ":all:", width: "auto" }
+
+      elem.attr "id", decko.uniqSelect2Id(elem.attr("id"))
```
xithan: You generate a unique "data-select2-id" id. Seems wiser to me to put that on "data-select2-id" (otherwise the next one won't be unique)

ethn: I think I tried something like that and it doesn't work. I would need to re-confirm this, but I think select2 was creating its own `data-select2-id` based on the `id` (even if the `data-select2-id` was explicitly set).

I don't understand your comment that "the next one won't be unique".

xithan: If `select2()` (in line 23) immediately sets the `data-select2-id` then it's fine. I thought if you don't set select2-id here then the next select2 in the loop (line 14) doesn't get a unique id.

If setting `data-select2-id` explicitly doesn't help then there are several places in the code where we should fix that.

ethn: I'm pretty sure it does set it.  But I should confirm both (a) that it does and (b) that setting data-select2-id doesn't work. I don't really like changing the id, but my conclusion was that I had to.

If I confirm that, I'll add comments to that effect.

ethn:
Your MessagePack question isn't showing up in the code now, so I'll answer it here:

I kept running into more and more weird encoding issues with MessagePack. I'm not entirely convinced it _can't_ work, but it was leading to really ugly breakage that would, of course, only show up in production.  Short version: MessagePack produces a binary, which when converted to a string would often cause problems.

ethn:
(fixed list card issue)

PR #413, opened 2019-10-07
Rails upgrade

xithan:


ethn:
I think we can close this.  The only thing that's different from master now is the test data.  And if the tests are working on master, we don't need it, right?

PR #414, opened 2019-10-11
Rails upgrade PLUS

ethn:


card/card.gemspec
```diff
@@ -46,16 +47,16 @@ Gem::Specification.new do |s|
 
 
     # files and images
-    #    ["carrierwave",                "~> 1.1"],
-    ["carrierwave",                 "1.2.2"],
-    ["fog",                           "2.1"], # can be removed after carrierwave upgrade
+    ["carrierwave",                 "2.0.2"],
+    # ["fog",                           "2.1"], # can be removed after carrierwave upgrade
```
xithan: Since the carrierwave upgrade seems to work we can get rid of that. We probably have to add fog-aws to Wikirate's Gemfile when we upgrade. We don't load all the ~20 fog cloud services by default anymore.

PR #424, opened 2019-12-02
pointer api harmonization

ethn:


card/mod/history/lib/card/action/admin.rb
```diff
@@ -8,17 +8,6 @@ def delete_cardless
...
-      # {Change change}
-      def delete_changeless
-        joins(
-          "LEFT JOIN card_changes "\
-          "ON card_changes.card_action_id = card_actions.id"
-        ).where(
-          "card_changes.id IS NULL"
-        ).delete_all
-      end
-
```
ethn: A changeless action used to be a data problem. It's now the default case, and we pretty much never want to delete them.  I don't think it's wise even to offer the api to do so.

card/mod/pointer/set/abstract/02_pointer/item_api.rb
```diff
@@ -143,24 +152,30 @@ def item_type
   opt.item_type
 end
 
-# #item_names helpers
+def item_strings args={}
```
ethn: item_strings can return things like `"~234"` and `":codename"`.

ethn: ...but it has the same filtering (limit/offset) api as item_names.

PR #426, opened 2019-12-09
avoid initializing new card in fetch

ethn:


card/lib/card/set/event.rb
```diff
@@ -96,10 +96,10 @@ def event_type
       end
 
       def define_standard_event_method method_name=simple_method_name
-        stage = @stage
+        is_integration = @stage.to_s.match?(/integrate/)
         @set_module.class_exec(@event) do |event_name|
           define_method event_name do
-            rescuing_integration stage do
+            rescuing_if_integration is_integration do
```
ethn: I was seeing some errors saying "stage not defined", but I didn't really have a good way to reproduce them.  So I'm not sure this will help, but it seems a little more efficient anyway...

card/mod/core/set/all/fetch.rb
```diff
@@ -124,23 +124,29 @@ def fetch opts={}
   end
 end
 
-def renew args={}
-  opts = args[:new].clone
+def renew new_opts
+  return self if new_opts.blank?
+  opts = new_opts.clone
+  copy = dup
```
ethn: the main change to fetch here is that renew makes a duplicate rather than instantiating a new object.  That way cached attributes are retained.

card/mod/core/set/all/name.rb
```diff
@@ -55,21 +55,25 @@ def key= newkey
 def update_subcard_names new_name, name_to_replace=nil
   return unless @subcards
   subcards.each do |subcard|
```
ethn: this is mostly cleanup. the change is _not_ altering the subcard name if the subcard is real.

PR #437, opened 2020-01-15
#first_name and card#fetch 

ethn:


card/mod/core/set/all/fetch.rb
```diff
@@ -115,8 +115,8 @@ def fetch_type_id *mark
 # fetching from the context of a card
 
 def fetch traits, opts={}
-  traits = Array.wrap traits
-  traits.inject(self) do |card, trait|
+  opts[:new][:supercard] = self if opts[:new]
```
ethn: I merged the earlier change already but should probably call more attention to it.

You no longer use `card.fetch trait: :traitname`.  Instead it's now just `card.fetch :traitname`.

PR #442, opened 2020-02-10
Markdown and history fixes

xithan:


card/lib/card/content/chunk.rb
```diff
@@ -89,6 +89,7 @@ def validate_chunk_list_key chunk_list_key
       register_list :nest_only, [:Nest]
       register_list :query, [:QueryReference]
       register_list :stub, [:ViewStub]
+      register_list :references_keep_escaping,  %i[KeepEscapedLiteral Nest Link]
```
ethn: I don't love this nomenclature, but I have now failed more than once to come up with anything better.

card/mod/edit/set/all/form_buttons.rb
```diff
@@ -5,7 +5,7 @@ def standard_submit_button
 
   def standard_save_button opts={}
     return if voo&.hide?(:save_button)
-    add_class opts, "submit-button btn-sm mr-3"
+    add_class opts, "submit-button btn-sm mr-3 _update-history-pills"
```
ethn: does this do something that should be covered in tests?

card/mod/core/chunk/keep_escaped_literal.rb
```diff
@@ -0,0 +1,26 @@
...
+          FULL_RE[prefix[1, 1]]
+        end
+
+        def interpret match, _content
+          @process_chunk = match[0].sub(/^\\(.)/, '\\\\\\\\\1')
+        end
+      end
+    end
+  end
+end
```
ethn: I think we have specs for every other chunk type, no?  Can we add a quick one for this one?

card/mod/history/set/all/history.rb
```diff
@@ -74,3 +94,10 @@ def save_content_draft content
        .card_changes.build(field: :db_content, value: content)
   end
 end
+
+event :update_timestamps, :finalize do
```
ethn: `update_ancestor_timestamps`, maybe?

card/mod/standard/spec/set/all/history_spec.rb
```diff
@@ -236,4 +236,18 @@
       end
     end
   end
+
+  describe "timestamping" do
```
ethn: add comment:

> \# note: B includes Z, so updates to Z stamp B

PR #443, opened 2020-02-10
try new autoloading approach

ethn:


decko/lib/decko/application.rb
```diff
@@ -20,16 +20,6 @@ class Application < Rails::Application
       end
     end
 
-    initializer :load_card, after: :load_config_initializers, group: :all do
```
ethn: It's all lazy loading now.

PR #460, opened 2020-04-04
fix machine cache bug

xithan:


card/mod/carrierwave/set/abstract/attachment.rb
```diff
@@ -20,7 +20,7 @@ def self.included host_class
   @current_action.update! comment: original_filename
 end
 
-event :validate_file_exist, :validate, on: :save do
+event :validate_file_exist, :validate, on: :create do
```
xithan, line 23: this is another unrelated problem: File cards that where created with :empty_ok cause problems when they are renamed. This solution assumes that once a card was created without a file updates without a file are ok.

card/mod/machines/set/abstract/machine/output_update.rb
```diff
@@ -1,8 +1,10 @@
 def reset_machine_output
   Auth.as_bot do
     moc = machine_output_card
+    @updated_at = output_updated_at
     moc.delete! if moc.real?
     update_input_card
+    expire_if_source_file_changed @updated_at
```
xithan, line 7: after the card is delete we can't get the updated_at date anymore. So we have to hold onto it here for

PR #475, opened 2020-07-22
WIP: new name handling

ethn:


card/spec/lib/card/query/card_query/match_attributes_spec.rb
```diff
@@ -12,7 +12,7 @@
     end
 
     it "gets only name when name is explicit" do
-      expect(run_query(name: [:match, "two"])).to eq(["One+Two", "One+Two+Three", "Two"])
+      expect(run_query(name: [:match, "two"])).to eq(["Two"])
```
xithan, line 15: Will you fix this?

PR #482, opened 2020-08-27
Remove captcha

GerryG:
This just takes out the mod code and removes it from Modfile. I will make another PR to card-mods if it isn't already there.


ethn:
This is a great start. We'll have to do a few more things to keep existing sites from breaking.

1. somehow make it so `decko update` manages optionally adding the mod gem to the Gemfile 
2. either abstract admin_info to be a general tool (I think that's my preference) or add a mechanism to handle all the old nests

You can probably use the other gemspecs in the card-mods repo as a template.  I think we may eventually have some auto-magic related to naming conventions, so it's worth trying to follow that pattern.

Thanks so much for working on this!

card/mod/Modfile
```diff
@@ -27,6 +27,5 @@ mod 'account'
 mod 'navbar'
 mod 'bootstrap'
 mod 'google_analytics'
-mod 'recaptcha'
```
ethn, line 30: I think if we change this to gem_mod we can plausibly merge this PR, though I'm concerned about the loss of CI testing of recaptcha.  We'd need to solve that...

ethn (changes requested):

card/mod/Modfile
GerryG: Yes.

I have the other branch with more mod/X components moved to gem_mod within the card repo. Maybe that's where recaptcha should be.

It is on in default Decko, yes? I'll let you decide the correct default configuration.
Technically you just add it as a Gemfile dependency. I couldn't decide whether these should or should not be in the Modfile. It can control load order, but if you want all the gem_mod/* at the end, should be fine without them in Modfile.

GerryG: Maybe reject this PR and I add recaptcha to the other change? I will post a PR for that since it seems to be mostly working now.

GerryG:
Superceded by mod2gems PR

PR #486, opened 2020-09-06
Mod2gems - move some mod/X Modules to gem_mod and make them gems

GerryG:
There was a bit of trickyness with finding stuff and loading right, but I think this is a good way to go. A good first step.


GerryG:
I just added recaptcha back, but in gem_mods with the other moved mods. This should be working. I will keep an eye on what is up with tests if they still fail.

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
```diff
@@ -0,0 +1,21 @@
+# -*- encoding : utf-8 -*-
+
+Gem::Specification.new do |s|
+  s.name = "card-mod-ace_editor"
+  s.version = "0.1"
+
+  s.authors = ["Ethan McCutchen", "Philipp Kühl"]
```
ethn, line 7: Please make yourself third author on these.

card/gem_mods/edit/set/all/editor.rb
```diff
@@ -46,6 +44,7 @@ def default_input_type
 
   # overridden by mods that provide rich text editors
   def rich_text_input
+    #FIXME: should select default differently, don't depend on an editor
     send "#{Cardio.config.rich_text_editor || :tinymce}_editor_input"
```
ethn, line 48: This should eventually probably say `Cardio.config.rich_text_editor || :text_area`, because `text_areat_input` is the simple input defined in this mod. But for that to work we'd have to update existing decks and add `config.rich_text_editor = :ace` to config/application.rb.  We should probably be gathering those cases so that we can do them all at once.

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
```diff
@@ -0,0 +1,21 @@
+# -*- encoding : utf-8 -*-
+
+Gem::Specification.new do |s|
+  s.name = "card-mod-ace_editor"
+  s.version = "0.1"
```
ethn, line 5: As long as this is in the same repo (and, more importantly, the same testing environment) as decko and card, I think we should probably make these versions use the same version as decko.  Eg `decko 0.10.1` must use `card-mod-edit 0.10.1`.  That's the easiest way to make sure they'll work together.  You can see what I'm talking about in decko.gemspec (although the card gem requires special handling because it predates decko and is already way past 1.0)

ethn (changes requested):
This is awesome and quite close.

We need to handle the abstract/all issue discussed in slack, and there are some other minor issues mentioned inline.

card/gem_mods/edit/set/all/editor.rb
GerryG: That should make it just a plain browser editor, no wysiwyg, right? (Done)

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
GerryG: Ok, I think I can just pull in the card version same as card.gemspec

GerryG:
Did I get them all?

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
GerryG: I think i missed one comment for completion, but this is done.

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
GerryG: Oh, this one. I will see if this is easy(ish)

GerryG: I hope that is right. I locked the versions together in the card.gemspec which means they all release together.

GerryG:
I guess we have to tell it how to build all the gems on semaphoreci? I have a script that I just used in my workspace, you want it?

GerryG:
I think I fixed it, but I accidentally pushed more changes to the branch. I started playing with breaking up mod/core into subparts. I'd like your mini-review on that separately.

GerryG:
Oops, didn't mean to close it. Oh, I deleted and re-added the branch. I guess I need a new PR and leave this one closed.

PR #490, opened 2020-09-15
Mostly removing wagn refs and changing to decko

GerryG:
There is also some updates to the docker directory to update it to adapt to changes in master (bit rot). This is not complete, but should not be used by anyone else.


ethn (changes requested):
NICE! 

Let's get this green and merge it in.

I think there are just two things that need to happen:

1. Go to db/seed/test/fixtures and fix the wagn references. I think if you change the @wagn.org to @decko.org, it should fix the spec.  While you're there, it's also generally safe to remove the deprecated `/wagn` from urls (eg http://decko.org/wagn/HTML can just be http://decko.org/HTML).

2. There are a couple styling issues to fix on codeclimate.  We're not always strict on those, but if they're easy to knock out, this is a nice mechanism for slowing getting the code cleaner over time (even if you weren't the one who originated the style violation).  Obviously not a dealbreaker.

GerryG:
I think I have that change as a commit I can add here. I'll get this, I would like to merge it sooner than later.

GerryG:
I think I have the fixture changes in, but maybe not on the branch. I must have regenerated them. I will regen with this branch to make sure they are right.

card/mod/search/set/abstract/wql_search.rb
```diff
@@ -43,23 +43,23 @@ def standardized_query_args args
   args
 end
 
-def wql_hash
-  @wql_hash = nil unless cache_query?
-  @wql_hash ||= wql_content.merge filter_and_sort_wql
+def cql_hash
```
ethn, line 46: some of these are overridden and/or called in non-core mods, including wikirate mods. We'll have to fix those, too.

ethn (changes requested):
What's our plan for the mods affected by these method changes?  Can we get in some PRs for them?

card/mod/search/set/abstract/wql_search.rb
GerryG: Just saw that. I'll take a quick look Is this the only method name I changed, or are there more?

PR #492, opened 2020-09-18
Mod2gems - move some mod/X Modules to gem_mod and make them gem 2

GerryG:
New PR because I broke the old branch


card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
```diff
@@ -0,0 +1,29 @@
...
+vbits = version.split('.').map &:to_i
+vplus = { 0 => 90, 1 => 100 } # can remove and hardcode after 1.0
+vminor = vplus[ vbits[0] ] + vbits[1]
+card_version = [1, vminor, vbits[2]].compact.map(&:to_s).join "."
+# see card.gemspec for explanation of all of this, which has been ham-handedly
+# cut and pasted here.
+
+Gem::Specification.new do |s|
+  s.name = "card-mod-ace_editor"
+  s.version = card_version
```
ethn, line 13: we actually just want "version", not card_version.  card_version is going to be weird until we get to decko 1.0.  Then every gem will just use the same version.

ethn (changes requested):
This PR is never going to pass (or even really run) CI tests until we take care of template/Gemfile. I think I already explained how to do that, but let me know if you need more help.

card/gem_mods/ace_editor/card-mod-ace_editor.gemspec
GerryG: Ok, I'll change that. I wasn't sure how that would work with the other version streams.

GerryG:
I'm not sure I get the Gemfile issue. I have to change the generator so it makes a good Gemfile in decko new. Is that the gist of it. I guess I need to use that directly in my Docker stuff too. That would test it before it gets to ci. I think I'll try that.

decko/decko.gemspec
```diff
@@ -12,7 +12,7 @@ card_version = [1, vminor, vbits[2]].compact.map(&:to_s).join "."
 
 Gem::Specification.new do |s|
   s.name          = "decko"
-  s.version       = version
+  s.version       = card_version
```
ethn, line 15: yikes! no not this.

GerryG:
I think I fixed it and the version change.

decko/decko.gemspec
GerryG: fixed all the versions

decko/lib/decko/generators/decko/templates/Gemfile
```diff
@@ -6,8 +6,24 @@ path "<%= @gemfile_gem_path %>" do
...
   gem 'cardname', :require => false
 end
+path "<%= @gemfile_gem_path %>/card/gem_mods" do
+  gem 'card-mod-date'
+  gem 'card-mod-recaptcha'
+  gem 'card-mod-ace_editor'
+  gem 'card-mod-prosemirror_editor'
+  gem 'card-mod-tinymce_editor'
+  gem 'card-mod-edit'
+  #gem 'card-mod-layout'
```
ethn, line 16: I think we can delete (not just comment) these.

ethn: (mod-layout only)

card/mod/Modfile
```diff
@@ -21,12 +18,15 @@ mod 'media'
...
 mod 'bootstrap'
 mod 'google_analytics'
-mod 'recaptcha'
+gem_mod 'recaptcha'
 gem_mod 'markdown'
+gem_mod 'date'
+gem_mod 'ace_editor'
+gem_mod 'prosemirror_editor'
+gem_mod 'tinymce_editor'
+gem_mod 'edit'
```
ethn, line 31: Based on what we learned the other day about how gem_mods are loaded, these may not be necessary?

GerryG: Yeah, not at the end anyway. I just moved them back to the original order.

.gitmodules
```diff
@@ -19,11 +19,11 @@
 [submodule "card/mod/machines/vendor/jquery_rails"]
 	path = card/mod/machines/vendor/jquery_rails
 	url = https://github.com/rails/jquery-rails
-[submodule "card/mod/date/vendor/moment"]
-	path = card/mod/date/vendor/moment
+[submodule "card/gem_mods/date/vendor/moment"]
```
ethn, line 22: realizing we should singularize `gem_mod` to be consistent with `mod`

ethn: also, this should be `card-mod-date`, not just `date`, right?

GerryG:
I think I know what is breaking. I'm not sure about what those tests are using for Gemfile, is that to make sure it will run on a previous configuration (Gemfile). It is supposed to work without referencing the card-mod-X gems?

PR #497, opened 2020-09-22
Card cli - CI test

GerryG:
I'm expecting this to break, but I want to see if the new tasks will still support the CI scripts as they are. Will fix when it makes sense, plan on updating the scripts if that's the better angle. We can pre-release some of the commands stuff if we want to do that.


GerryG:
Closing to stop CI on commits. Will open when it works more.

PR #499, opened 2020-09-23
card-mod-defaults

ethn:


card/gem_mod/card-mod-ace_editor/card-mod-ace_editor.gemspec
```diff
@@ -20,4 +20,5 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.3.0"
   s.metadata = { "card-mod" => "ace_editor" }
   s.add_runtime_dependency "card-mod-edit", Versioning.simple
+  s.add_runtime_dependency "card", Versioning.card
```
ethn, line 23: I guess this is redundant...

card/mod/Modfile
```diff
@@ -23,10 +23,5 @@ mod 'navbar'
 mod 'google_analytics'
 gem_mod 'edit'
```
GerryG, line 24: We need the Modfile to put this before bootstrap? Seems like a small issue we should fix when code gets moved around and into more gems. then this ordering is in the default pack gem.

card/gem_mod/card-mod-ace_editor/card-mod-ace_editor.gemspec
GerryG: right, if you depend on another card-mod-gem, it likely already has card in the chain of deps. So either on card or another card-mod- is enough.

GerryG:
If it builds and tests right, all the specs, etc. then it looks good.

card/mod/Modfile
ethn: There is no need to worry about ordering in the pack.

`card-mod-bootstrap` will depend on `card-mod-edit`. Then worry not.

card/gem_mod/card-mod-ace_editor/card-mod-ace_editor.gemspec
ethn: fixed

PR #507, opened 2020-09-26
Card cli - restarted the branch

GerryG:
To run CI


card/lib/card/config/environments/test.rb
```diff
@@ -1,5 +1,5 @@
 # -*- encoding : utf-8 -*-
-Decko.application.class.configure do
+Card::Application.configure do
```
GerryG, line 2: This variation loads, but even though it looks like seed -t is loading fixtures, no test data (load of joe user fails in early test setup). I think this pattern works.

ethn (commented):
A lot of good work here. Tomorrow I'll write PK and get him to weigh in.

ethn:
Is this branch still viable?  Do we keep the PR open?

ethn:
I think we've moved on to other branches, right?

PR #508, opened 2020-09-26
Making new gemspec stuff work for installing gems

GerryG:
Include as module instead of using inheritance


ethn:
My solution's better :)

PR #517, opened 2020-10-01
moving more mods (history, accounts, etc) into mods.

ethn:


ethn:
trying to trigger semaphore.

PR #518, opened 2020-10-02
disallow broken version of "better errors"

ethn:


ethn:
this was a fine PR. then I screwed it up.

PR #561, opened 2020-10-20
Cli2simple

ethn:


card-mod-carrierwave/set/abstract/attachment/cloud.rb
```diff
@@ -34,7 +34,7 @@ def bucket_config
 
 def load_bucket_config
   return {} unless bucket
-  bucket_config = Cardio.config.file_buckets&.dig(bucket.to_sym) || {}
+  bucket_config = ::Cardio.config.file_buckets&.dig(bucket.to_sym) || {}
```
ethn, line 37: these shouldn't be necessary unless there is a subclass called Cardio

card-mod-history/set/all/history/events.rb
```diff
@@ -32,7 +32,11 @@ def edit_conflict?
 # removes the action if there are no changes
 event :finalize_action, :finalize, when: :finalize_action? do
   if changed_fields.present?
-    @current_action.update! card_id: id
+    if @current_action.new_record?
+      @current_action.card_id = id
+    else
+      @current_action.update! card_id: id
+    end
```
ethn, line 39: why are you messing with this?

GerryG: Mistake, reverted

.semaphore/semaphore.yml
```diff
@@ -28,7 +28,8 @@ blocks:
         commands:
           - cache restore bundle-gems-baseline
           # occasionally clear: `cache delete bundle-gems-baseline`
-          - cache restore bundle-gems-$(checksum Gemfile.lock)
+          - cache delete bundle-gems-$(checksum Gemfile.lock)
```
ethn, line 31: you shouldn't ever need to do this.  just do `bundle update`

card/lib/card/cache.rb
```diff
@@ -106,7 +106,7 @@ def reset_soft
       # (the non-standard caches)
       def reset_other
         Card::Codename.reset_cache
-        Cardio.delete_tmp_files!
+        ::Cardio.delete_tmp_files!
```
ethn, line 109: these are the ones I don't see value in.

decko/decko.gemspec
```diff
@@ -9,7 +9,7 @@ DeckoGem.gem "decko" do |s, d|
 
   s.files = Dir["{db,lib,public,set}/**/*"]
 
-  s.bindir = "bin"
+  s.bindir = "exe"
```
ethn, line 12: Why?

decko/lib/decko/application.rb
```diff
@@ -1,90 +1,69 @@
...
-        # (c) CAN be overridden in an environment file
+    initializer :deck_autoload, before: :set_autoload_paths do |app|
+      Cardio.set_config
+      ActiveSupport::Dependencies.autoload_paths += config.autoload_paths
 
-        # therefore, in general, they should be restricted to settings that
-        # (1) are specific to the web environment, and
-        # (2) should not be overridden
-        # ..and we should address (c) above!
+      # any config settings below:
```
ethn, line 33: this no longer makes grammatical sense

decko/lib/decko/application.rb
```diff
@@ -1,90 +1,69 @@
...
-        config.i18n.enforce_available_locales = true
-        # config.active_record.raise_in_transactional_callbacks = true
+      # general card settings (overridable and not) should be in cardio.rb
+      # overridable decko-specific settings don't have a place yet
+      # but should probably follow the cardio pattern.
 
-        config.allow_concurrency = false
-        config.assets.enabled = false
-        config.assets.version = "1.0"
+      config.load_defaults "6.0"    # note this isn't a setter method
```
ethn, line 47: this could have a lot of weird effects, no?

card/lib/card/format.rb
```diff
@@ -72,7 +72,7 @@ def params
     end
 
     def controller
-      @controller || Env[:controller] ||= CardController.new
+      @controller || Env[:controller] ||= ::CardController.new
```
ethn, line 75: `::` is pretty much never needed before Cardio or CardController, because there are no subclasses named Cardio or CardController.  The point of this `::` is to distinguish, eg, `Cardio` from `Card::Format::Cardio`. Since nothing like the latter exists, it's just noise here.

card/lib/cardio.rb
```diff
@@ -105,24 +126,22 @@ def default_configs
...
-      @@config = config
-
-      add_lib_dirs_to_autoload_paths config
+    def set_config
+      add_lib_dirs_to_autoload_paths
 
-      default_configs.each_pair do |setting, value|
-        set_default_value(config, setting, *value)
+      default_configs.each_pair do |setting, *value|
+        config.send("#{setting}=", *value) unless config.respond_to? setting
```
ethn, line 133: You lost the comment that explains why we do this. Is it no longer applicable?

card/lib/tasks/card/migrate.rake
```diff
@@ -94,7 +94,7 @@ namespace :card do
     desc "write the version to a file (not usually called directly)"
     task :stamp, :type do |_t, args|
       ENV["SCHEMA"] ||= "#{Cardio.gem_root}/db/schema.rb"
-      Cardio.config.action_mailer.perform_deliveries = false
+      Card.config.action_mailer.perform_deliveries = false
```
ethn, line 97: I don't think Card is already loaded here, so Cardio may make more sense.

ethn: That's the rule of thumb for deciding which to use.  We want to avoid loading Card to get the config object.

decko/lib/decko/commands/rake_command.rb
```diff
@@ -16,26 +16,30 @@ def initialize rake_task, args={}
...
+      def command
+        @command ||= command_with_env
       end
 
-      def commands
-        task_cmd = "bundle exec rake #{@task}"
-        return [task_cmd] if !@envs || @envs.empty?
-        @envs.map do |env|
+      def command_with_env
+        @envs.inject(task_cmd) do |task_cmd, env|
```
ethn, line 33: Doesn't this significantly change the logic?

It seems like the old way actually ran the command multiple times. This looks like it makes one command with multiple environments, which doesn't make sense.

decko/lib/decko/engine.rb
```diff
@@ -15,43 +15,54 @@
...
-    paths["lib/tasks"] << "#{::Cardio.gem_root}/lib/card/tasks"
-    paths.add "lib/decko/config/initializers",
-              with: File.join(Decko.gem_root, "lib/decko/config/initializers"),
-              glob: "**/*.rb"
 
-    initializer "decko.engine.load_config_initializers",
-                after: :load_config_initializers do
-      paths["lib/decko/config/initializers"].existent.sort.each do |initializer|
-        load(initializer)
+    engine_name = :decko_engine
```
ethn, line 19: I'd have to study more to understand exactly what you've done here, but it sure looks a lot cleaner!

ethn (commented):
Since I made the PR, I can't actually approve or request changes, but I think the comments inline cover things.

This looks like a very strong step in the right direction for us.  Thank you!!

.semaphore/semaphore.yml
GerryG: Yeah, I just don't understand the whys and whens of these processes yet. I'm pretty sure I restored to master with this config.

card/lib/card/cache.rb
GerryG: I think I have the loading of the CardClassMethods on Card slightly wrong. I think it should do 'Card.extend ...' in Cardio where it loads mods on the load hook.

It's like Card::Cardio gets defined empty and catches these. I think this is in Card.rb, but not sure yet.

decko/decko.gemspec
GerryG: This is the new standard in rails. This is to avoid confusion with bin (old way was binstubs). For a 'bin', something in this dir (exe in the gem, not in files list, this is infra stuff), it creates a bin in gems/bin and in the system (/usr/local/bin, I think) for the entry point that loads bundler and the command from the gem. That stub calls this file. Equivalent is 'exe' for rails now. This stuff does not need the script loading, this is the place for that stuff.

decko/lib/decko/application.rb
GerryG: Thanks for finding what was breaking with this. I rolled it back, got stuff to work and now just re-added with that exception overriden. I think this is good if it doesn't break stuff, yes?

ethn (commented):
Please go back to the green and get rid of all this new stuff.

ethn:
Closing because this is inactive and now has substantial conflicts with the master branch.  Feel free to repair and reopen.

PR #689, opened 2021-06-08
supplement to xithan's assets pull

ethn:
Cardio::Mod was a module with all class methods
Card::Mod was a mod with all instance methods.

There is now just Cardio::Mod, a class with class methods and instance methods.

And it means cukes can run.

(cukes were failing because Card wouldn't autoload. It's not enough to require the application to get autoload going. For that you have to actually _initialize_ the application)


ethn:
xithan merged

PR #701, opened 2021-07-27
Getting wikirate working with new assets handling

ethn:


mod/assets/set/abstract/mod_assets.rb
```diff
@@ -130,14 +130,15 @@ def ensure_item_save card, args
   if card
     card.update args
   else
-    Card.create args.merge(name: item_name)
+    Card.create! args
```
ethn, line 133: Not sure where this "item_name" was coming from, but I don't think it's what we wanted.

mod/script/set/type/local_script_manifest_group.rb
```diff
@@ -1,2 +1,2 @@
-include_set Abstract::LocalManifestGroup
 include_set Abstract::ScriptAssetList
+include_set Abstract::LocalManifestGroup
```
ethn, line 2: Order is important here. Otherwise minimize? is always true.

decko/lib/decko/tasks/decko.rake
```diff
@@ -94,6 +94,7 @@ decko_namespace = namespace :decko do
       decko_namespace["migrate"].invoke
       decko_namespace["reset_tmp"].invoke
       Card::Cache.reset_all
+      decko_namespace["mod_install"].invoke
```
ethn, line 97: always run `rake decko:mod_install` on `decko update`.  Incentive to optimize...

mod/assets/set/abstract/asset_list.rb
```diff
@@ -57,13 +57,8 @@ def local
   @local = true
 end
 
-def minimize?
-  @minimize = true
-end
-
```
ethn, line 63: this was overriding the manifest settings

PR #807, opened 2022-02-06
nest editor overhaul

xithan:


mod/assets/set/abstract/asset_inputter.rb
```diff
@@ -34,6 +34,7 @@ def update_asset_input
 
   Card::Auth.as_bot do
     asset_input_card.update content: render_asset_input_content
+    Card::Cache.reset_all
```
xithan, line 37: Do we have something better than this to make sure that `asset_input` returns the updated content from the line above?

PR #817, opened 2022-02-24
rollback default sorting for now

ethn:


ethn:
this caused problems when used as a subquery. reject.

PR #823, opened 2022-03-10
asset fixes

ethn:


mod/assets/set/abstract/asset_inputter.rb
```diff
@@ -33,7 +33,7 @@ def update_asset_input
   # otherwise the migration that adds the asset_input card fails
 
   Card::Auth.as_bot do
-    asset_input_card.update content: render_asset_input_content
```
ethn, line 36: This was leading to double rendering, because the (virtual) asset input card was not using this content but rather its #virtual_content method (which also calls render_asset_input_content on this card).

mod/virtual/lib/card/virtual.rb
```diff
@@ -19,11 +19,13 @@ def fetch card
 
       def save card
         virt = find_by_card card
-        virt ? virt.update(card.virtual_content) : create(card)
+        virt ? virt.update(card.virtual_content) : (virt = create card)
+        cache.write card.key, virt
```
ethn, line 23: doh. not updating cache

xithan (approved):

mod/virtual/lib/card/virtual.rb
xithan: this already fixed but differently in #807

ethn: ah, right. I prefer yours.

xithan (approved):

PR #824, opened 2022-03-11
recent changes optimization

ethn:


mod/script/assets/script/decko/modal.js.coffee
```diff
@@ -15,6 +15,10 @@ $(window).ready ->
   $('body').on 'click', '.submit-modal', ->
     $(this).closest('.modal-content').find('form').submit()
 
+  $('body').on 'click', '._close-modal-now', ->
+    $(this).closest('.modal').modal "hide"
```
xithan, line 19: Is this different from what bootstrap does when you use `class="close" data-dismiss="modal"`?

ethn: probably not. will go with that.

PR #835, opened 2022-04-11
Bootswatch inheritance

ethn:


mod/bootstrap/set/type/bootswatch_skin.rb
```diff
@@ -1,15 +1,174 @@
...
+#
+# For the original bootswatch themes all those parts are hard-coded and the content
+# is taken from files.
+# The bootswatch theme content is taken directly from the files in the bootswatch
+# submodule. For the rest we use code file cards.
+# Cards of type "customized bootswatch skin" have the same structure but make the
+# variables and stylesheets part editable.
+#
+# Bootswatch theme cards are "asset inputters," meaning they generate a
+# Machine cards usually store all involved input cards of all nested levels in
```
xithan, line 27: old wording - machines are gone

mod/assets/lib/card/assets.rb
```diff
@@ -30,6 +30,15 @@ def refresh_assets force: false
         generate_asset_output_files if force
       end
 
+      def wipe
+        Auth.as_bot do
+          %i[asset_input asset_output].each do |field|
+            Card.search(right: field).each(&:delete!)
+            Virtual.where(right_id: field.card_id).destroy_all
```
xithan, line 37: Did this do the trick?

ethn: It works, but this isn't what solved the problem. The issue was in the variables card and happened even when re-eating from a fresh deck.  Still, nice to have around.

mod/bootstrap/set/type/bootswatch_skin.rb
ethn: Right. I started in on this edit but never finished it. Thanks.

PR #841, opened 2022-04-18
ymlizing

ethn:


card/lib/cardio/commands.rb
```diff
@@ -20,7 +20,7 @@ def gem
     def map
       @map ||= {
         new: { desc: "create a new deck", group: :shark, via: :call },
-        seed: { desc: "populate a database", group: :shark, via: :rake },
+        setup: { desc: "populate a database", group: :shark, via: :rake },
         update: { desc: "run data updates", group: :shark, alias: :u, via: :rake },
```
ethn, line 24: Worth noting and discussing if there is concern. The new seeding approach means we can go back to supporting a lot more standard rails rake tasks:

- db:seed 
- db:seed:replant (truncate tables and seed)
- db:setup (create and seed)
- db:reset (drop and setup)

What we were calling "seed" is a lot more like what rails calls "setup", so I thought maybe we should embrace that pattern.  Will need to update more docs if we go forward with this.

card/lib/cardio/seed.rb
```diff
@@ -6,15 +6,87 @@ module Seed
...
-        parts = [Cardio.gem_root, "db"] + args
-        File.join(*parts)
+      def path
+        if ENV["UPDATE_SEED"]
+          args = Rails.env.test? ? ["production", 0] : ["production", 1]
+          db_path(*args)
+        else
+          default_path
+        end
+      end
```
ethn, line 21: There's a lot of power in this little bit of code. Let me give some context.

So there are two kinds of seed yaml files:
- fixtures, which are the standard rails things that mimic data tables directly, and
- pods, or whatever we're calling our special seed representation that mimics our API

Fixtures are generally a pain to write by hand. You have to manage ids, actions, references, etc.  So there is only one place we're doing that manually for now, which is in `card/mod/core/data/fixtures` . That defines our current minimal seeding of 11 cards.

All other seed data will be generated by (a) reseeding with existing features, and then (b) eating pods to add to it. But, importantly, this process is nestable. So the production seed data in the defaults mod uses the core fixtures, the test seed data in that mod uses the production data, etc.  Wikirate seed data will use fixtures from the defaults mod, of course.

These lines of codes are really what make that whole system tick (and will probably need to get more sophisticated)

card/mod/core/lib/tasks/card/seed.rake
```diff
@@ -0,0 +1,38 @@
+namespace :card do
+  namespace :seed do
+    desc "reseed, migrate, re-clean, and re-dump"
+    task update: :environment do
+      ENV["STAMP_MIGRATIONS"] = "true"
```
ethn, line 5: in concept, this same task should work for wikirate seeding

card/spec/card/query/card_query/sorting_spec.rb
```diff
@@ -24,16 +24,16 @@
...
+  # it "sorts by plus card content" do
+  #   Card::Auth.as_bot do
+  #     Card["Setting+*self+*table of contents"].update! content: 10
+  #     Card.create! name: "RichText+*type+*table of contents", content: "3"
+  #     expect(run_query(right_plus: "*table of contents",
+  #                      sort: { right: "*table_of_contents" },
+  #                      sort_as: "integer"))
+  #       .to eq(%w[*all RichText+*type Setting+*self])
+  #   end
+  # end
```
ethn, line 36: I probably need to rewrite these tests before merging. They were failing because the cards aren't in the seed data any more.

xithan (approved):
LGTM

PR #881, opened 2022-06-28
Fix image autocomplete; validate script and style

xithan:


mod/style/set/type/css.rb
```diff
@@ -1,2 +1,18 @@
...
+  #validator = W3CValidators::CSSValidator.new
+  #results = validator.validate_text(content)
+  #results.errors.each do |error|
+  #  errors.add :content, "#{error.message} (line #{error.line})"
+  #end
+  begin
+    ::Sass::SCSS::Parser.new(content, nil, nil).parse
+  rescue ::Sass::SyntaxError => e
+    errors.add :content, e.message
+  end
```
xithan, line 17: Css validation is harder than expected. 
Our Scss-Parser (SassC) can't validate plain css. The commented solution above (W3CValidator) needs a web request for the validation. The solution I picked uses a deprecated Scss-Parser.

PR #968, opened 2023-02-21
fixing input type on dates

vasgat:
add type="date" attr on text_field

### DIFF
--- a/mod/date/set/all/calendar.rb
+++ b/mod/date/set/all/calendar.rb
@@ -8,7 +8,8 @@ format :html do
 
   def calendar_input
     text_field :content, class: "date-editor datetimepicker-input",
-                         "data-bs-toggle": "datetimepicker",
-                         "data-bs-target": "##{form_prefix}_content.date-editor"
+               "type": "date",
+               "data-bs-toggle": "datetimepicker",
+               "data-bs-target": "##{form_prefix}_content.date-editor"
   end
 end


mod/date/set/all/calendar.rb
```diff
@@ -8,7 +8,8 @@ def datepicker_config
 
   def calendar_input
     text_field :content, class: "date-editor datetimepicker-input",
-                         "data-bs-toggle": "datetimepicker",
-                         "data-bs-target": "##{form_prefix}_content.date-editor"
+               "type": "date",
```
ethn, line 11: so the datetimepicker stuff is supposed to hook this up to a special js library (but is evidently failing to do so). What does "type":"date" do? Is that like an HTML5 thing? Do we think it has pretty good browser support?

ethn (changes requested):
Don't worry about the failing cucumber tests; that has something to do with an updated gem that we don't manage. It even broke main. I think I have a fix coming.

But please to fix the minor codeclimate issues.

More importantly, I want to be clear on what our solution is. If we're moving away from the old date-picker, I want to make sure that we're doing so thoughtfully. What is the date representation stored with the new handling? Is it compatible with the old one? Are we going to need to migrate date data?

mod/date/set/all/calendar.rb
vasgat: Hmmm, I am not sure what I have done is a good solution. The type="date" just says that the input field on the html is of type date and it actually allows the user to pick a date from a calendar. I just saw that you are using a js script for the date picker called tempusdominus (found under vendor). However, the tempusdominus does not support bootstrap 5 only bootstrap 3 and 4 and this might be the reason the datepicker is broken. 

I think we could solve this by using bootstrap 5 and html with type=date (some migration on the database might indeed be needed (I am not sure in what type of format old versions of date are stored).

ethn: Hey, nice sleuthing on the bootstrap 5 issue. That makes a lot of sense.

It looks to me like the old format (or the default at least) was YYYY-MM-DD. Or, at least, that's what the cards look like here: https://wikirate.org/Date+*type+by_name?item=content

But maybe it's already messy. The datepicker config card makes it look like it's supposed to be just YY rather than YYYY.  Odd: https://wikirate.org/:datepicker

In any case, support for the new way of doing things seems strong, and we can also implement several of the other types: https://caniuse.com/input-datetime

This is almost certainly  more than you hoped to undertake at this point. So maybe if we can confirm that the new editor defaults to YYYY-MM-DD andwe fix the codeclimate issues and I write up a PT story in the decko project and merge in this PR.

PR #975, opened 2023-03-05
Improve transitions from content-only editor to board and back

xithan:


mod/edit/set/all/bridge.rb
```diff
@@ -65,4 +67,22 @@ def breadcrumb_data title, html_class=nil
...
+  def bridge_menu
+    wrap_with_modal_menu do
+      [
+        close_modal_window,
+        switch_to_edit_link
+      ]
+    end
+  end
+
+  def switch_to_edit_link
```
ethn, line 80: I'd like to get more consistent about making helper methods like these private.

mod/format/set/all/html/menu.rb
```diff
@@ -146,4 +144,9 @@ def modal_icon
   def bridge_icon
     icon_tag :board
   end
+
+  def confirm_edit_loss opts
```
ethn, line 148: (also private)

ethn (approved):

PR #976, opened 2023-03-05
Override slot options for bridge overlays

xithan:


mod/format/set/all/html/views.rb
```diff
@@ -3,23 +3,12 @@
 format :html do
   def prepare_content_slot
     class_up "card-slot", "d0-card-content"
-    voo.hide :menu
+    voo.hide :menu, :edit_button
```
ethn, line 6: This is definitely an improvement.

Arguably, the more appropriate API for this is `optional: :hide`, eg:
```
wrap do
  [
    render_menu(optional: :hide), 
    render_core, 
    render_edit_button(optional: :hide, edit: :inline)
  ]
end
```

Also, not necessarily for this release, but it kind of seems like we could get rid of `content_with_title`, too, using a show/hide approach.

mod/format/assets/script/decko/slotter.js.coffee
```diff
@@ -201,8 +203,8 @@ jQuery.fn.extend
     slotter.addClass "_close-#{type}"
 
   slotterBeforeSend: (opt) ->
-    return if opt.noSlotParams
-
     # avoiding duplication. could be better test?
-    unless (opt.url.match(/home_view/) or @data("slotter-mode") == "modal")
-      opt.url = decko.slot.path opt.url, @slot()
+    return if opt.noSlotParams or opt.url.match(/home_view/) or
+      @data("slotter-mode") == "modal" or @data("slotter-mode") == "override"
```
ethn, line 208: this is a little cognitively demanding.

ethn (changes requested):
This looks good to go to me once the spec is fixed.

mod/format/assets/script/decko/slotter.js.coffee
xithan: I thought three "or"s  are less cognitively demanding than a mixture of "if" and "unless"

mod/format/set/all/html/views.rb
xithan: That changes the view cache key. I'd rather don't mess with the view cache.

mod/format/assets/script/decko/slotter.js.coffee
ethn: neither are very good though, right?  might benefit from its own function.

PR #1014, opened 2023-10-05
Login signup

cschweipert:
Toggles visibility of password with eye icon and adds password validation.


mod/account/set/right/password.rb
```diff
@@ -38,9 +38,16 @@ def ok_to_read
...
+
+  def password_input
+    haml :password_input
+  end
+
   view :input do
-    card.content = ""
-    password_field :content, class: "d0-card-content", autocomplete: autocomplete?
+    password_input
   end
```
ethn, line 51: *Either* you want to override the input view, *or* you want to use the `input_type` pattern. But not both. (The second option is probably preferable)

The thing that calls `input_type` is the input view. If you're overriding the input_view, then there is no need for `input_type`.

mod/account/set/right/password/password_input.haml
```diff
@@ -0,0 +1,3 @@
+%input.form-control.d0-card-content#card_subcards___password_content{name: "card[subcards][+*password][content]", type: "password", placeholder: "Password"}
```
ethn, line 1: This has the word `*password` hardcoded into it. That means if someone changes the name of the `*password` card, it will break.

If you change it from `+*password` to `+:password`, however, it should be fine.

I also think we can do without the weird id (`#card_subcards___password_content`)

ethn: This would be a good time to bring up i18n (internationalization). Can you please use the configuration in `mod/account/config/locales` to make the "Password" placeholder language-specific?

mod/bootstrap/set/all/bootstrap/icon.rb
```diff
@@ -32,7 +32,9 @@
     reset: :restart_alt,
     more: :more_horiz,
     menu: :more_vert,
-    bar_menu: :more_horiz
+    bar_menu: :more_horiz,
+    eye_open: :visibility,
+    eye_slash: :visibility_off
```
ethn, line 37: I would think it would be more intuitive to use something like `show_password` and `hide_password` for our keys here. More importantly, we want to make sure that we use these keys in the code. (You have used the material name rather than the decko name in the code).

Also, we generally want to try to make it so that we have all the icon keys mapped in font awesome, too, so that someone can switch over to font awesome without things breaking. Can you add these new ones in font awesome as well?

ethn (changes requested):

mod/account/set/right/password/password_input.haml
cschweipert: Where is this weird id coming from? It's the id of the input field by default.
Changing *password to :password breaks it (screenshot).
Screenshot 2023-10-11 at 3 19 18 PM

mod/account/set/right/password.rb
cschweipert: I am not sure I understand. The input view automatically calls input_type? I don't see input_type called anywhere. Is this a naming convention? name_of_type gets called by view :name_of?

ethn (changes requested):
Am I missing something? I don't see code that enforces password requirements at all.

mod/account/assets/script/toggle_visiblity.js.coffee
```diff
@@ -0,0 +1,11 @@
+$ ->
+  passwordField = $("input._toggle-password")
+  togglePassword = $("span._toggle-password")
+
+  togglePassword.on "click", () ->
```
ethn, line 5: This is the issue, not the "open" view. The problem is when the view is loaded via ajax. I guarantee it would happen with any view.

So if you use this code, then when the document is ready (`$ ->`) it will look for all `toggle-password` spans and attach click events to them. That's great if the the span is there when the document loads. It's not great if the span doesn't appear until later via an ajax call. In that case, nothing happens to attach a click event to those spans.

We encountered this before with the attribution. The solution we used there was something like `$("body").on "click", togglePassword`. In this case the click event is attached to the body tag. It's something like "when you click on the body, check to see if you're clicking inside a password field.".  That approach works even when you use ajax.

ethn (changes requested):

mod/account/assets/script/toggle_visibility.js.coffee
```diff
@@ -0,0 +1,12 @@
+$ ->
+  $("body").on "click", "._toggle-pw-visibility", (event) ->
+    passwordField = $("._pw-input")
+    togglePassword = $("._pw-text-area")
+
+    if passwordField.prop("type") is "password"
+      passwordField.prop("type", "text")
```
ethn, line 7: not a big deal, but it's slightly more idiomatic not to use parentheses here. We're not using any style rules on coffeescript for now. Maybe someday soon?

mod/account/set/right/password.rb
```diff
@@ -2,6 +2,13 @@
 
 assign_type :phrase
 
+PASSWORD_REGEX_REQ = {
+  lower: [/[a-z]/, :account_password_requirement_lower],
+  upper: [/[A-Z]/, :account_password_requirement_upper],
+  special_char: [/[!@#$%^&*()]/, :account_password_requirement_special_char],
+  number: [/\d+/, :account_password_requirement_number]
+}.freeze
```
ethn, line 10: Why not just do, eg
```
{ lower: /[a-z]/ }
```

You can construct the i18n key out of the hash key.

cschweipert:
codeclimate:
1. complains about variable shadowing but it is necessary to pass pw_requirements into the each_with_object block in order for it to work. 
2. it complained about using single quotes here before I changed it to double quotes.

mod/account/set/right/password.rb
```diff
@@ -15,14 +22,45 @@ def ok_to_read
...
+  min_pw_length = Cardio.config.account_password_length
+  return if content.length >= min_pw_length
+
+  errors.add :password, t(:account_password_length, num_char: min_pw_length)
+end
+
+def check_password_regex char_types, regex_hash, password
+  pw_requirements = char_types.each_with_object([]) do |char_type, pw_requirements|
+    if regex_hash.key?(char_type) && password !~ regex_hash[char_type][0]
+      pw_requirements << regex_hash[char_type][1]
```
ethn, line 39: could just do `:"account_password_requirement#{char_type}"`, right?

mod/account/set/right/password.rb
```diff
@@ -15,14 +22,45 @@ def ok_to_read
...
+    end
+  end
+  pw_requirements unless pw_requirements.empty?
+end
+
+def format_error_message err_messages
```
ethn, line 45: might be better to have a narrower name. if you have a card and see that it has a method "format_error_message", you wouldn't assume it only applies to one specific error message.

maybe something like `consolidated_password_error_message`?

mod/account/set/right/password.rb
```diff
@@ -15,14 +22,45 @@ def ok_to_read
...
+  error_message = err_messages.map { |message| t(message) }
+  if error_message.length > 2
+    "#{error_message[0...-1].join(", ")}, and #{error_message[-1]}"
+  else
+    error_message.join(" and ")
+  end
+end
+
+event :validate_password_chars, :validate, on: :save do
```
ethn, line 54: Would you be wiling to put the events at the top of the file and make the methods that are called by the events (but only the events) private?

mod/account/spec/set/right/password_spec.rb
```diff
@@ -17,14 +17,13 @@ def card_subject
...
       expect(password).not_to eq("new password")
-      authenticated = Card::Auth.authenticate "joe@user.com", "new password"
+      authenticated = Card::Auth.authenticate "joe@user.com", "new Pas5word!"
       expect(account).to eq authenticated
     end
 
     it "validates password" do
-      password_card = account.password_card
       password_card.update content: "2b"
       expect(password_card.errors[:password]).not_to be_empty
```
ethn, line 28: I would like a little more sophisticated testing here. Can you show that the length and character requirements are both working? I would rather we not just test for the presence of errors but also that the errors are correct.

ethn (changes requested):
Nice work! This is very close. The main remaining work:

- code climate changes: they look easy
- simplification of the regexp handling. It's definitely heading in the right direction, but I think the code can be a little simpler/cleaner
- testing of the character requirement stuff

cschweipert:
> Nice work! This is very close. The main remaining work:
> 
> * code climate changes: they look easy
> * simplification of the regexp handling. It's definitely heading in the right direction, but I think the code can be a little simpler/cleaner
> * testing of the character requirement stuff

mod/account/set/right/password.rb
```diff
@@ -15,14 +22,45 @@ def ok_to_read
...
-  return if content.length > 3
+event :validate_password_length, :validate, on: :save do
+  min_pw_length = Cardio.config.account_password_length
+  return if content.length >= min_pw_length
+
+  errors.add :password, t(:account_password_length, num_char: min_pw_length)
+end
+
+def check_password_regex char_types, regex_hash, password
+  pw_requirements = char_types.each_with_object([]) do |char_type, pw_requirements|
```
ethn, line 37: what codeclimate doesn't like is that you're using the same variable name for the array inside the iteration and the one outside. That makes it harder to understand what the code is doing.

mod/account/set/right/password.rb
```diff
@@ -15,14 +22,45 @@ def ok_to_read
...
+  end
+  pw_requirements unless pw_requirements.empty?
+end
+
+def format_error_message err_messages
+  error_message = err_messages.map { |message| t(message) }
+  if error_message.length > 2
+    "#{error_message[0...-1].join(", ")}, and #{error_message[-1]}"
+  else
+    error_message.join(" and ")
```
ethn, line 50: previously you had single quotes here; that's what the previous code climate issue was about.  The ones above (line 48 right after the join) should be single because their inside an interpolated string.

ethn (changes requested):

mod/account/set/right/password.rb
cschweipert: It is intended to be the same variable as in the outer scope.

mod/account/set/right/password.rb
cschweipert: I am not sure what you mean.

mod/account/set/right/password.rb
ethn: it's the same value, but it's two different variables.

cschweipert:
I think we can ignore the line too long code climate issue

mod/account/set/all/password_input.haml
```diff
@@ -0,0 +1,3 @@
+%input.form-control.d0-card-content._pw-input{name: "card[subcards][+*password][content]", type: "password", placeholder: "Password"}
```
ethn, line 1: We don't want the cardname ("*password") hardcoded here. That would mean that if someone renamed the password card, the signin would break.  Can we use ":password" instead? The colon indicates a codename, which will not change even if the cardname does.

cschweipert: Did you see my above comment? It breaks when I change it. See above with screenshot.

PR #1017, opened 2023-10-23
admin.yml API

xithan:
Each mod can provide in an admin.yml file a list of features that it adds to the deck.

The currently supported categories are:
- cardtypes
- settings
- tasks
- configurations
- views
- styles
- scripts

TODO: We need a less confusing view for settings in the settings section on role cards.

PR #1019, opened 2023-11-04
Fix style rule editor

xithan:
Based on #1017 because of the Cypress upgrade that is included there.
Fixes only one Coffeescript file and adds a Cypress tests


ethn (commented):
It's quite possible I'm being thick, but I don't actually see any code changes that seem like they could fix the style rule editor (except, I guess, gem updates?). I see some good tests and some styling fixes. 

What fixed things?

xithan:
The `window.ready` in `click_select.js.coffee` is the fix

xithan:
I don't understand the broken cypress tests. It looks like the four failing tests have no css.

xithan:
I have a theory now: Maybe the "getting start" cyp breaks the css.

SEEDME.md
```diff
@@ -64,7 +64,7 @@ core
 
 This tells you:
 
-1. Running `be decko setup` will use fixtures from the **defaults** mod, the first mod on
+1. Running `bundle execute decko setup` will use fixtures from the **defaults** mod, the first mod on
```
ethn, line 67: exec, not execute.

ethn (approved):

PR #1023, opened 2023-11-29
Pw vis req

cschweipert:
This adds the letter requirement to the general pw requirement and modifies the special characters to spec.


mod/account/lib/card/mod/account.rb
```diff
@@ -1,4 +1,4 @@
 Cardio::Railtie.config.tap do |config|
   config.account_password_length = 8
-  config.account_password_requirements = %i[lower upper special_char number]
+  config.account_password_requirements = %i[lower upper special_char number letter]
```
ethn, line 3: I'm not sure we need both "letter" _and_ "lower"/"upper".  I think in this context we just drop letter.

mod/account/set/right/password.rb
```diff
@@ -5,8 +5,9 @@
 PASSWORD_REGEX = {
   lower: /[a-z]/,
   upper: /[A-Z]/,
-  special_char: /[!@#$%^&*()]/,
-  number: /\d+/
+  special_char: /[!§$%&#]/,
```
ethn, line 8: No objection, but interested to hear what inspired this change?

ethn (changes requested):
only tiny tweaks requested. Very close!

mod/account/set/right/password.rb
ethn: Ah, I see. I see that she used the word "like".  I don't think she meant to imply that we would use _only_ those characters.  This would be a pretty weird, random set. I think she meant to give some examples even though we might support a lot more.

Is that third character non-ascii?  This discussion gives some decent reasons for not using unicode (https://security.stackexchange.com/questions/225346/is-it-bad-to-use-special-characters-in-passwords). And I certainly don't think we should use only one unicode character.

In this case I think perhaps we revert and discuss with Melanie.

cschweipert: It might be good to add this link to pt so she can look at it before the discussion.

mod/account/spec/set/right/password_spec.rb
```diff
@@ -57,6 +57,15 @@ def card_subject
       expect(password_card.errors[:password]).to eq []
     end
 
+    it "works with 'letter' requirement" do
+      Cardio.with_config account_password_requirements: [:letter] do
+        password_card.update content: "1234567890"
+        expect(password_card.errors[:password]).to eq [
+          "must contain a letter"
+        ]
```
ethn, line 65: nice!

ethn (approved):

PR #1026, opened 2023-12-04
Comments

cschweipert:
Some method "doc strings" that help me and maybe somebody else.


mod/bootstrap/set/all/bootstrap/navbar.rb
```diff
@@ -40,6 +40,10 @@ def navbar_responsive id, content, opts, nav_opts
     end
   end
 
+  # Generates HTML markup for a breadcrumb trail.
+  # @param items [list]: A list of items representing the breadcrumb trail.
+  # Returns:
+  # str: HTML markup for the breadcrumb trail.
```
ethn, line 46: Yardoc (which we use to generate, eg docs.decko.org) understands @param (yay!) and @return

```
@return [String] HTML markup...
```

mod/edit/set/all/form_buttons.rb
```diff
@@ -3,6 +3,18 @@ def standard_submit_button
     output [standard_save_button, standard_save_and_close_button]
   end
 
+  # Generates a standard save button with optional customization.
+  # @param opts (dict, optional): A dictionary containing customization options
```
ethn, line 7: "dict" is Hash in rubyspeak.

mod/edit/set/all/form_buttons.rb
```diff
@@ -3,6 +3,18 @@ def standard_submit_button
...
+  #   - If voo is not nil and does not hide the save button, a save button is generated
+  #   - If present, "text" key in opts sets the text content of the button
+  #   - "data-cy" attribute is set to "save" for end-to-end testing purposes
+  #   - CSS class "submit-button me-3 _update-history-pills" is applied
+  # Returns:
```
ethn, line 12: See @options handling here: https://gist.github.com/chetan/1827484

mod/edit/set/all/form_buttons.rb
```diff
@@ -3,6 +3,18 @@ def standard_submit_button
...
+  #   - None: If the save button is not generated due to the hide condition.
+  # Example:
+  #   standard_save_button({"text": "Submit", "data-cy": "submit_button"})
+  #   This example generates a customized save button with the text "Submit"
```
ethn, line 16: you wouldn't ever set "data-cy" to anything; it will just get overwritten. `:text`, on the other hand, used `||=`, so you can set that as an option. All other options are passed on to `#submit_button`

ethn (changes requested):
Thanks for working on this!

Please see @return and @option conventions (links in other comments)

mod/edit/set/all/form_buttons.rb
```diff
@@ -3,6 +3,19 @@ def standard_submit_button
     output [standard_save_button, standard_save_and_close_button]
   end
 
+# Generates a standard save button with optional parameters.
+#
+# @param [Hash] opts The options for the save button.
+# @option opts [String] :text ("Save") The text displayed on the button.
+# @option opts [String] :data-cy ("save") The value for the 'data-cy' attribute
+#   used for testing or identification purposes.
```
ethn, line 11: this option is not correct.  It's pretty much the only thing that _won't_ get passed on to `submit_button`

mod/edit/set/all/form_buttons.rb
```diff
@@ -3,6 +3,19 @@ def standard_submit_button
...
+#
+# @return [String] The HTML code for the standard save button.
+#
+# @example
+#   standard_save_button(text: "Update")
+#   #=> "<button class='submit-button me-3 _update-history-pills' data-cy='save'\
+#   data-disable-with='Submitting'>Update</button>"
```
ethn, line 18: here's a good example of why long explicit examples cause problems. If someone changes the default "disable-with" value in `submit_button`, it would be very hard for them to realize that some other documentation somewhere assumed it wouldn't change.

Fragile documentation is even worse than no documentation, because it ends up being incorrect and therefore worse than having to read the code.

mod/format/set/all/links.rb
```diff
@@ -26,9 +26,11 @@ def link_to_view view, text=nil, opts={}
   end
 
   # link to a card other than the current card.
-  # @param cardish [Integer, Symbol, String, Card] a card identifier
-  # @param text [String]
-  # @param opts [Hash]
+  # @param cardish [Integer, Symbol, String, Card]: A card identifier
+  # @param text [String], optional: The text content of the link.
+  # @param opts [Hash], optional: Additional options for the link.
+  # Returns:
```
ethn, line 32: @return

mod/format/set/all/links.rb
```diff
@@ -79,6 +81,13 @@ def clean_resource resource, resource_type
     end
   end
 
+  # Adds key-value pairs from a new hash to the 'path' key
+  # in the given options dictionary.
+  # @params opts [dict]: The options dictionary to be modified.
+  # @params new_hash [dict]: The new hash containing
+  # key-value pairs to be added to the 'path'.
+  # Returns:
+  #   None
```
ethn, line 90: @return

(None is incorrect; it actually does return the value of `opts[:path]`)

ethn (changes requested):

PR #1027, opened 2023-12-05
Tooltips

cschweipert:
I added the title and reverted the tooltips. Let's wait for Melanie's response to see if this is an acceptable solution for her style wise. The titles for the other menu icons are going to be in a separate wr PR.


ethn (approved):
Looks good!

PR #1030, opened 2023-12-08
fix password bug

ethn:


mod/account/set/right/password.rb
```diff
@@ -5,7 +5,7 @@
 PASSWORD_REGEX = {
   lower: /[a-z]/,
   upper: /[A-Z]/,
-  special_char: /[!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~]/,
```
ethn, line 8: @cschweipert I'm a little confused about how this passed before, but the main branch is busted. That must mean I approved and merged a branch that broke things. 

This seems to fix things; just wanted you to know.

cschweipert: @ethn Oh. Thanks for fixing it.

cschweipert (approved):

mod/account/set/right/password.rb
```diff
@@ -5,7 +5,8 @@
...
+  # special_char: /[!"#$%&'()*+,\-\.\/:;<=>?@\[\]\\^_`{|}~]/,
+  special_char: /[!"#$%&'()*+]/,
```
ethn, line 9: Actually, my solution wasn't working. I don't have time to look at this right now. @cschweipert: Can you see if you can get a version that (a) works, (b) passes code climate, and (c) passes semaphore tests?

For now I just deleted a bunch of characters so that the main branch won't be broken.

PR #1034, opened 2023-12-19
a few codename-rather-than-id-constant fixes

ethn:


xithan:
Is the ID constant approach deprecated now?

ethn:
I wouldn't say deprecated, no. There are places where it has given me problems, though. (Mostly when the constant is called during loading). And in CQL I definitely find the codename syntax cleaner.

PR #1039, opened 2024-01-25
update carrierwave

ethn:


mod/carrierwave/lib/carrier_wave/card_mount.rb
```diff
@@ -29,13 +29,13 @@ def mount_uploader column, uploader=nil, options={}, &block
...
-        end
+        # event :mark_remove_#{column}_false_event, :finalize, on: :update do
+        #   mark_remove_#{column}_false
+        # end
 
-        event :store_previous_changes_for_#{column}_event, :store,
+        event :reset_previous_changes_for_#{column}_event, :store,
               on: :update, when: proc { |c| !c.history? } do
-          store_previous_changes_for_#{column}
+          reset_previous_changes_for_#{column}
```
ethn, line 38: There is not longer a "store_previous_changes" method in mount.rb. Maybe they renamed it to this because that's more accurate? Or maybe it's something different?

mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -175,6 +175,10 @@ def url_filename
       end
     end
 
+    # def temporary_identifier
+    #   db_content
+    # end
```
ethn, line 180: There were several issues caused by these "temporary identifiers". It has to do with what gets written to the `mount_on` field (in our case, `#db_content`). This fixed certain problems but caused several others.

mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -232,16 +236,16 @@ def storage
...
+    # def content_type
+    #   # the original content_type method doesn't seem to be very reliable
+    #   # It uses mime_magic_content_type  - which returns invalid/invalid for css files
+    #   # that start with a comment - as the second option.  (we switch the order and
+    #   # use it as the third option)
+    #   @content_type ||=
+    #     # existing_content_type ||
+    #     mini_mime_content_type ||
+    #     mime_magic_content_type
+    # end
```
ethn, line 248: This method completely broke, as none of the methods it called exist any more. But I'm not very confident that the correct content types are still found.

mod/carrierwave/set/abstract/attachment/coded.rb
```diff
@@ -29,13 +29,17 @@ def storage_type_error error_name
 end
 
 def mod_from_content
-  if (m = content&.match %r{^:[^/]+/([^.]+)})
+  if (m = mod_name_match content) # || mod_name_match(db_content_before_act))
```
ethn, line 32: The remaining failing spec is connected to the `#write_temporary_identifier` method in mounter.rb in carrierwave. (not a typo; mounter not mount). I tried a hacky workaround, but it broke stuff.

mod/carrierwave/lib/carrier_wave/card_mount.rb
```diff
@@ -29,13 +29,13 @@ def mount_uploader column, uploader=nil, options={}, &block
           remove_#{column}!
         end
 
-        event :mark_remove_#{column}_false_event, :finalize, on: :update do
-          mark_remove_#{column}_false
-        end
+        # event :mark_remove_#{column}_false_event, :finalize, on: :update do
+        #   mark_remove_#{column}_false
+        # end
```
ethn, line 34: I don't know how important this is. This event was triggering the `#write_temporary_identifier` thing, which was overwriting the correct `#db_content_value` with the incorrect one. (You can reproduce by uncommenting and running `coded_spec.rb`)

I worry this omission could have unintended side effects.

Gemfile.lock
```diff
@@ -73,9 +73,9 @@ PATH
       card (= 1.106.0)
       card-mod-history (= 0.16.0)
       card-mod-permissions (= 0.16.0)
-      carrierwave (~> 2.2)
-      mini_magick (~> 4.2)
-      ssrf_filter (~> 1.0.7)
+      carrierwave (~> 3.0)
```
ethn, line 76: This is the big gem change.

mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
xithan: mini_mine was replaced with marcel https://github.com/carrierwaveuploader/carrierwave/pull/2552

mod/carrierwave/lib/carrier_wave/card_mount.rb
xithan: It's something different: https://github.com/carrierwaveuploader/carrierwave/commit/aac25c10af4218d6e1e70f90154b847b54ce0334

mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
xithan: This is the new approach: https://github.com/carrierwaveuploader/carrierwave/blob/d27c9871c83613717bd69d31da7d23450bf266c9/lib/carrierwave/sanitized_file.rb#L241
Maybe it works now and we don't have to override it anymore.

mod/carrierwave/lib/carrier_wave/card_mount.rb
xithan: I think just removing breaking code is not a good approach. This section should be adjusted to the changes in the carrierwave code. What we do here is replacing the activerecord hooks in https://github.com/carrierwaveuploader/carrierwave/blob/d27c9871c83613717bd69d31da7d23450bf266c9/lib/carrierwave/orm/activerecord.rb#L29 with card events

mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -231,19 +235,6 @@ def storage
...
-      # It uses mime_magic_content_type  - which returns invalid/invalid for css files
-      # that start with a comment - as the second option.  (we switch the order and
-      # use it as the third option)
-      @content_type ||=
-        existing_content_type ||
-        mini_mime_content_type ||
-        mime_magic_content_type
-    end
-  end
-
```
xithan, line 246: Maybe we have to reorder the new mime type recognitions methods here:
```
        identified_content_type ||
        declared_content_type ||
        guessed_safe_content_type ||
        Marcel::MimeType::BINARY
```

`identified` returns nil for wikirate's css, `guessed` returns `plain/css`. I don't know what `declared` returns

ethn: Yeah, I realized as we drove away that this must be the problem area.

PR #1105, opened 2024-09-17
Fix argument handling for "sow" and "eat"

xithan:


xithan:
Close because Codeclimate is confused

PR #1107, opened 2024-09-17
Fix argument handling for rake tasks "sow" and "eat"

xithan:


card/card.gemspec
```diff
@@ -3,6 +3,7 @@
 require "../decko_gem"
 
 DeckoGem.gem "card" do |s, d|
+  required_ruby_version = ">= 3.0.0"
```
ethn, line 6: DeckoGem handles required_ruby_version

xithan: Codeclimate doesn't like it if a gem does not specify a ruby version.

card/card.gemspec
```diff
@@ -3,6 +3,7 @@
 require "../decko_gem"
 
 DeckoGem.gem "card" do |s, d|
+  s.required_ruby_version = ">= 3.0.0"
```
ethn, line 6: We should get rid of these and fix CodeClimate.  DeckoGem handles this, and imo it's preferable that we only maintain this in one place.

ethn: (by "fix" I mean tell it to exclude that check)

PR #1110, opened 2024-09-23
try newing action rather than creating it

ethn:


mod/carrierwave/set/type/file.rb
```diff
@@ -9,7 +9,8 @@ def select_action_by_params params
 
   def last_content_action_id
     # find action id from content (saves lookups)
-    db_content.to_s.split(%r{[/.]})[-2]
+    aid = db_content.to_s.split(%r{[/.]})[-2]
+    aid.number? ? aid.to_i : nil
```
ethn, line 13: this was at least temporarily a codename when changing from coded to local storage.

card/lib/card/director/class_methods.rb
```diff
@@ -25,7 +25,7 @@ def run_act card
       end
 
       def need_act
-        self.act ||= Card::Act.create ip_address: Env.ip
+        self.act ||= Card::Act.new ip_address: Env.ip
```
ethn, line 28: does not create until needed

mod/history/set/all/history/events.rb
```diff
@@ -81,6 +67,26 @@ def edit_conflict?
 
 private
 
+def new_action
+  Card::Action.new(
```
ethn, line 71: does not create until needed

PR #1250, opened 2025-12-09
fix: WDM-6 return .jsonld hydra:next and :previous links instead of .…

vasgat:
…json


ethn (changes requested):
Would you be willing to try the suggested fix?

